PR #80, opened 2024-03-31
Use levenshtein score to approximate moves.

mattmoor:
One of the most common sources of diffs we are seeing comparing the filesystems of APKs is when there is an ABI change, but not an actual capability change.  This shows up as a potentially noisy Deleted / Added with the full capability listing.

This change adds a post-pass to the reporting computation that walks the Added/Deleted file lists, and if two paths are a >90% match, then we treat them as a "move" and combine their reports in a similar fashion to modifications.

This also incorporates the notion of moves into the report and adjusts some of the rendering to surface these as moves with the similarity score (so we can tune this).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,10 +3,12 @@ module github.com/chainguard-dev/bincapz
 go 1.21.6
 
 require (
+	github.com/agext/levenshtein v1.2.3
 	github.com/google/go-cmp v0.6.0
 	github.com/hillu/go-yara/v4 v4.3.2
 	github.com/liamg/magic v0.0.1
 	github.com/olekukonko/tablewriter v0.0.5
+	golang.org/x/term v0.18.0
 	gopkg.in/yaml.v3 v3.0.1
 	k8s.io/klog/v2 v2.120.1
 )
@@ -15,5 +17,4 @@ require (
 	github.com/go-logr/logr v1.4.1 // indirect
 	github.com/mattn/go-runewidth v0.0.9 // indirect
 	golang.org/x/sys v0.18.0 // indirect
-	golang.org/x/term v0.18.0 // indirect
 )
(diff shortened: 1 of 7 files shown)


mattmoor:
Here's an example of what `wolfictl check diff` surfaces with this change (see bottom for move):

<details>
  <summary>bincapz found differences: Click to expand/collapse</summary>

## Deleted: xz-dev/usr/share/doc/xz/examples_old/xz_pipe_comp.c

|   RISK    |        KEY        |                           DESCRIPTION                            |
|-----------|-------------------|------------------------------------------------------------------|
| -2/MEDIUM | ref/path/relative | references and possibly executes relative path: "./xz_pipe_comp" |

## Deleted: xz-dev/usr/share/doc/xz/examples_old/xz_pipe_decomp.c

|   RISK    |        KEY        |                            DESCRIPTION                             |
|-----------|-------------------|--------------------------------------------------------------------|
| -2/MEDIUM | ref/path/relative | references and possibly executes relative path: "./xz_pipe_decomp" |

## Added: xz-doc/usr/share/man/man1/xzgrep.1

Overall risk: ✅ 1/LOW

|  RISK  |          KEY          |                   DESCRIPTION                    |
|--------|-----------------------|--------------------------------------------------|
| +1/LOW | **compression/bzip2** | works with bzip2 files                           |
| +1/LOW | **compression/gzip**  | works with gzip files                            |
| +1/LOW | **compression/zstd**  | zstandard - fast real-time compression algorithm |

## Moved: xz/usr/lib/liblzma.so.5.4.6 -> xz/usr/lib/liblzma.so.5.6.1 (score: 0.955556)


</details>

mattmoor:
FWIW, I am not attached to this in any way (e.g. if we think there's a cleaner way of implementing this), but I do think we need to find a way to make reviewing moves a bit nicer.

mattmoor:
Another example of the problem this is solving for: 
image

tstromberg (approved):
Sounds good to me. The only alternative I have to offer is to constrain the distance diff to numerical values only, but lets try this approach first. Thanks!

Issue #75, opened 2024-03-27
`go install` triggers malware detection


josephlewis42:
Installing `bincapz` as specified in the README using `go install` causes go to download and save the testdata directory in its mod cache. `testdata` contains samples that trigger malware scanners.

Would it be possible to either distribute binaries,  make it so the samples are obfuscated (e.g. in zip files with the "infected" password or encrypted), or something else to prevent this?

Thanks!

tstromberg:
I think that should be possible - any tip as to what scanner was triggering this?

cruiserh:
The tool uses a scanner developed by ReversingLabs.

tstromberg:
I asked because I've run into something similar with Elastic Defend—but it triggers on the YARA rules rather than the testdata.

I'm open to providing binaries if someone contributes the GitHub actions configuration for doing so.

PR #110, opened 2024-04-11
Refactor so that testdata samples are in their own namespace

tstromberg:
Related to #75 

This was particularly nice as we were able to move tests into their own namespace. 

We may move `bincapz.go` to `pkg/bincapz`, but I'm not sure yet.


cpanato:
need a rebase

pkg/report/report.go
```diff
@@ -297,16 +297,20 @@ func Generate(ctx context.Context, path string, mrs yara.MatchRules, ignoreTags
 	overallRiskScore := 0
 	riskCounts := map[int]int{}
 
-	for _, m := range mrs {
+	for x, m := range mrs {
+		clog.InfoContextf(ctx, "yara match[%d]: %+v", x, m)
```
vaikas, line 301: This is fine, but a better approach might be to take advantage of the structured logging fields, so we'd set the MatchRule once, and we don't have to keep logging it. This makes it easier to search for logs in some places because the "context" fields are set in the right context, and you don't need to keep tacking that in to each log message.
So, something like this here:
mrLog := clog.FromContext(ctx).With(slog.String("matchrule", m.Namespace))

And then below you could just do:
mrLog.Infof("dropping (risk %d is too low)", risk)
and
mrLog.Infof("key=%s", key)

so you don't have to keep logging the m.Namespace in different places. But just more FYI in case you weren't aware of it.

vaikas: In the other PR I did try to use it in places, but just wanted to take the first pass of moving to structured logging, I'm sure we can clean up the logging more intentionally later on.

vaikas (approved):

PR #117, opened 2024-04-11
Reduce noisy logging messages

tstromberg:
Now that log messages are plumbed through in tests, it's difficult to distinguish the important messages from the unimportant ones.

This removes logs that haven't been useful in a while, and reduces the logging level for others that may still come in handy. It also nukes some dead code (`skipFiles`) that is no longer necessary now that we are using YARA Forge.


vaikas (approved):
Yeah, was pretty verbose for sure :) And with info level too, so nice cleanup.

PR #123, opened 2024-04-16
Improve detection of embedded ZStandard content & high entropy binaries

tstromberg:
Noticed while investigating https://bazaar.abuse.ch/sample/e70e96983734ee23e52391aa96d30670b2dcebb0cbca46c8eddb014f450c661f/ 

It's a `sbcl` binary with a malicious ZStandard encrypted payload.


rules/evasion/packer/high_entropy.yara
```diff
@@ -0,0 +1,34 @@
+import "math"
+
+rule high_entropy_7 : suspicious {
```
vaikas, line 3: since the only thing different between these 4 rules is the math.entropy > 'number', I assume there's no way to reuse the entire other line, or somehow compose these? Here we match elf or macho, with filesize > 4096, but above it looks pretty much otherwise identical, but the filesize < 52428800.

I guess I'd expect that there should be something like this:
```
  condition $isElf and math.entropy(1, filesize) > 7.0
```

vaikas: time passes... I try to learn...

Kind of like here, but for Windows PE:
https://github.com/godaddy/yara-rules/blob/master/template.yara

Hm, looks like there's a way to maybe do something like I want above?

https://github.com/godaddy/yara-rules/blob/master/utility/IsElfFile.yara

tstromberg: What I was trying to do isn't possible yet anyways in bincapz, so I've removed the extra rules.

The TL;DR is, if multiple rules are matched in a single file, bincapz will only show the highest severity match. Getting the result I want will first require introducing a more advanced concept of rule precedence.

PR #125, opened 2024-04-16
Add GoReleaser Workflow

egibs:
Closes: https://github.com/chainguard-dev/bincapz/issues/75

This PR adds a Workflow to release binaries for `bincapz` which will add the artifacts to each release once a tag is created.

I included GPG signing configuration for the binaries as well (which means that `secrets.GPG_KEY` and `secrets.PASSPHRASE` will need to be added as repository secrets for this to work).


tstromberg:
The only thing I don't like here is the GPG integration - can we save that for another future PR?

The headache of maintaining long-term secrets isn't worth it, given how few people are likely to use it. I'd rather use `cosign` in the action, so at least there is a cryptographic guarantee that the binary was signed by the appropriate Github action, and not someone who pilfered the key.

egibs:
> The only thing I don't like here is the GPG integration - can we save that for another future PR?
> 
> The headache of maintaining long-term secrets isn't worth it, given how few people are likely to use it. I'd rather use `cosign` in the action, so at least there is a cryptographic guarantee that the binary was signed by the appropriate Github action, and not someone who pilfered the key.

Very fair point. I'll remove the GPG config for now and the signing can be a future PR 👍🏻

.github/workflows/release.yaml
```diff
@@ -0,0 +1,30 @@
+name: bincapz GoReleaser
+
+on:
+  push:
+    tags:
+      - '*'
```
tstromberg, line 6: make this: `- 'v*'`

tstromberg (changes requested):
Looks great! One minor letter change to make sure we only run this on version tags.

.github/workflows/release.yaml
egibs: Addressed in [`4899d0c` (#125)](https://github.com/chainguard-dev/bincapz/pull/12!5/commits/4899d0c62ce1df7bc4595f7255432820ffe55c7a)!

tstromberg (approved):

PR #126, opened 2024-04-17
python/shell rule improvements based on UPSTYLE analysis

### LINKED ISSUE #119: Improve detection for UPSTYLE
- https://bazaar.abuse.ch/browse/tag/UPSTYLE/
- https://www.volexity.com/blog/2024/04/12/zero-day-exploitation-of-unauthenticated-remote-code-execution-vulnerability-in-globalprotect-cve-2024-3400/

Here's the current state of output for one of the Python samples:

```
Path: 3de2a4392b8715bad070b2ae12243f166ead37830f7c6d24e778985927f9caac.unknown
Risk: 🚨 4/CRITICAL
--------------------------------------------------------------------------------------------
1/LOW   encoding/base64        supports base64 encoded strings
1/LOW   fs/file/delete         deletes files
2/MED   evasion/base64/decode  decodes base64 strings: "b64decode::$b64decode"
2/MED   fs/file/times/set      change file last access and modification times
4/CRIT  evasion/base64/python  contains base64 Python code: "leGVjK::$exec yZWFkKC::$read"
```

While that isn't bad, there are several signals we can incorporate, particularly the nearness of exec() to b64decode().

tstromberg:
Catches some simple backdoor techniques we might see in future supply-chain attacks.

closes #119


tstromberg:
closes #119

rules/combo/dropper/shell.yara
```diff
@@ -43,12 +43,24 @@ rule sleep_rm : notable {
...
 	$ref = /nohup bash [\%\w\/\>]{0,64} &/
   condition:
     any of them
 }
 
+rule fetch_pipe_shell_value : suspicious {
+  meta:
+	description = "fetches content and pipes it to a shell"
+  strings:
+	$wget_bash = /wget .{8,128}\| {0,2}bash/
```
vaikas, line 57: Just to make sure that I understand what this is saying. Any character after wget as long as it's between 8-128 in length? Why are we capping it at 128? And I guess conversely, why starting at 8? This wouldn't catch something like:
wget foo.io

Or am I misunderstanding something?

Also, I can't quite grok what the `{0,2}` is matching.

tstromberg: For performance reasons, YARA rules discourage unbounded regular expressions: https://github.com/Neo23x0/YARA-Performance-Guidelines/

But yeah, this says that any wget command line greater than 128 characters won't match. Otherwise, the wget and bash could potentially be 1GB apart.

I started at 8 because I figured you would have to specify the protocol, but to pipe to a shell,l you'll still need to provide `-O- foo.io` at a minimum, which is still 10  characters. 

the {0,2} will match 0 to 2 spaces.

vaikas:
```
export F='-O- example.com'
wget $F
```

Would it catch that?

Also, for curl below it's going to stdout by default, and don't need a protocol.

vaikas: Also, since generally the url limit is 2048, maybe bump to that?

tstromberg: No, it won't catch the `export` case or other obfuscations. 

To keep the scope small, the philosophy I've been going with for bincapz is to catch behaviors we've seen in the wild, and that's one I haven't seen yet. I have seen people rename curl though!

If I go up to 2048 for a URL, it would increase false-positive rates as the "." will include random binary content. I have not seen any malware samples where folks embed overly large URLs - it's more trouble than it's worth. 128 is already more than twice what has been observed.

vaikas: Ok, I'm just curious how much we want to try to be proactive vs. reactive here. Seems like renaming files might be another file that would catch these. So with the combination of rules, I'd expect things to be more tightened up. But, I trust you here :)

vaikas (approved):

PR #124, opened 2024-04-16
Add support for OCI images

egibs:
Closes: https://github.com/chainguard-dev/bincapz/issues/59

This PR adds the ability to scan OCI images by:
1. Pulling the image
2. Exporting the image to a temporary directory and `.tar` file
3. Extracting the `.tar` file
4. Scanning the temporary directory's files
5. Removing the temporary directory after the scan completes

I tested these changes out with the following images:
- `cgr.dev/chainguard/nginx`
- `cgr.dev/chainguard/wolfi-base`
- `python`

Here's the output from a Wolfi scan:
```sh
❯ go run . --oci cgr.dev/chainguard/wolfi-base
Path: /var/folders/3g/88131l9j11x995ppjbxsvhbh0000gn/T/bincapz-wolfi-base4241937653/bin/busybox
Risk: 🔥 3/HIGH
---------------------------------------------------------------------------------------------------------------------------------------------------
1/LOW   compression/bzip2           works with bzip2 files
1/LOW   compression/gzip            works with gzip files
1/LOW   encoding/base64             supports base64 encoded strings
1/LOW   env/HOME                    looks up the HOME directory for the current user
1/LOW   env/LANG                    looks up language of current user
1/LOW   env/LD_LIBRARY_PATH         ld library path
1/LOW   env/SHELL                   users preferred SHELL path
1/LOW   env/TERM                    look up or override terminal settings
1/LOW   env/TMPDIR                  tMPDIR
1/LOW   env/USER                    uSER
1/LOW   exec/program/background     waits for a process to exit
1/LOW   fs/directory/create         creates directories
1/LOW   fs/directory/remove         uses libc functions to remove directories
1/LOW   fs/fifo/create              make a FIFO special file (a named pipe)
1/LOW   fs/file/delete              deletes files
1/LOW   fs/file/truncate            truncate a file to a specified length
1/LOW   fs/link/read                read value of a symbolic link
1/LOW   fs/lock/update              apply or remove an advisory lock on a file
1/LOW   fs/mount                    mounts file systems
1/LOW   fs/node/create              able to make device files using mknod
1/LOW   fs/permission/chown         may change file ownership
1/LOW   fs/symlink/resolve          resolves symbolic links
1/LOW   fs/tempdir/create           uses libc functions to create a temporary directory
1/LOW   fs/tempfile/create          uses mktemp to create temporary files
1/LOW   fs/watch                    monitors filesystem events
1/LOW   kernel/cpu/info             gets number of processors
[...]

PR #144, opened 2024-04-17
Add support for archives

egibs:
This PR adds support for archives -- specifically `.apk`, `.jar`, `.tar`, `.tgz`, `.tar.gz`, `.tar.xz`, and `.zip` archives.

`.apk` and `.tar.*` files will use a `tar`-specific function and `.jar` and `.zip` files use a `zip`-specific function.

This behaves similarly to the code introduced in #124 where the archive is copied to a temporary directory and extracted before being scanned. I moved the `extract` code to `archive.go` and updated it to support `.tar.gz`, `.tar.xz` (via the `xz` package), and `.apk` files.

I tested this out with all of the newly-supported file types and `make lint` runs cleanly with these changes.

Examples are attached here:
[examples.md](https://github.com/chainguard-dev/bincapz/files/15012306/examples.md)


pkg/action/archive.go
```diff
@@ -0,0 +1,206 @@
...
+	"log"
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
+
+	"github.com/ulikunitz/xz"
+)
+
+const maxBytes = 1 << 29 // 512MB
```
egibs, line 19: This seems like an acceptably high archive size. If anything, it can probably be lower.

pkg/action/archive.go
```diff
@@ -0,0 +1,206 @@
...
+	"errors"
+	"fmt"
+	"io"
+	"log"
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
+
+	"github.com/ulikunitz/xz"
```
egibs, line 16: This was needed to support `.xz` files.

pkg/action/archive.go
```diff
@@ -0,0 +1,206 @@
...
+
+	if err := copyArchive(p, tmpDir); err != nil {
+		os.RemoveAll(tmpDir)
+		return "", fmt.Errorf("failed to copy archive: %w", err)
+	}
+
+	return tmpDir, nil
+}
+
+func extractTar(d string, f string) error {
```
egibs, line 60: This is essentially the `extract` function from #124 but modified to allow the other archive types.

pkg/action/oci.go
```diff
@@ -1,22 +1,17 @@
...
-	"strings"
 
 	"github.com/google/go-containerregistry/pkg/crane"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 )
 
-const maxBytes = 1 << 26
-
-func prepare(d string) (string, *os.File, error) {
+func prepareImage(d string) (string, *os.File, error) {
```
egibs, line 14: Small rename to be more OCI-specific.

bincapz.go
```diff
@@ -32,6 +32,7 @@ func main() {
 	statsFlag := flag.Bool("stats", false, "show statistics about the scan")
 	verboseFlag := flag.Bool("verbose", false, "emit verbose logging messages to stderr")
 	ociFlag := flag.Bool("oci", false, "scan an OCI image")
+	archiveFlag := flag.Bool("archive", false, "scan an archive (.zip, .tar, .jar, etc.)")
```
tstromberg, line 35: rather than introducing another flag, can we process archives transparently?

Ideally, I'd like bincapz to be able to process archives found in a recursive directory search, but this adds complication - so IMHO it's best to defer that to a second PR.

If it helps, there is already a `programKind` function that can be extended.

tstromberg:
Looks great! My only feedback is to see if we add this feature without introducing more flags for people to learn.

bincapz.go
egibs: Addressed this in [`c72990f` (#144)](https://github.com/chainguard-dev/bincapz/pull/144/commits/c72990f25f276ca936c4d764f7886dc8cac3cf0c) and [`3a0f288` (#144)](https://github.com/chainguard-dev/bincapz/pull/144/commits/3a0f288429c28797a5b00ab9aad65f5a7bd54302).

I added functionality to extract the file extension and used a separate map for the archive types to directly check whether the path is an archive or not.

pkg/action/archive.go
```diff
@@ -0,0 +1,225 @@
...
+	"path/filepath"
+	"strings"
+
+	"github.com/ulikunitz/xz"
+)
+
+const maxBytes = 1 << 29 // 512MB
+
+// copyArchive copies the source archive file to the temporary directory.
+func copyArchive(src string, dst string) error {
```
vaikas, line 22: We might want to plumb context here so that we get the correct logger.

pkg/action/archive.go
```diff
@@ -0,0 +1,225 @@
...
+	}()
+
+	if _, err = io.Copy(w, r); err != nil {
+		return fmt.Errorf("failed to copy data: %w", err)
+	}
+
+	return nil
+}
+
+// tempDir creates a temporary directory.
```
vaikas, line 47: This comment is a bit misleading since it not only creates the directory, but also copies the archive to it.

pkg/action/archive.go
```diff
@@ -0,0 +1,225 @@
...
+	defer r.Close()
+
+	w, err := os.CreateTemp(dst, fmt.Sprintf("%s.*", filepath.Base(src)))
+	if err != nil {
+		return fmt.Errorf("failed to create temporary file: %w", err)
+	}
+
+	defer func() {
+		if cerr := w.Close(); cerr != nil {
+			log.Printf("failed to close file: %v", cerr)
```
vaikas, line 36: I think this should be an error.

pkg/action/programkind.go
```diff
@@ -16,31 +16,40 @@ import (
 	"github.com/liamg/magic"
 )
 
+var archiveMap = map[string]bool{
+	".apk":    true,
+	".jar":    true,
+	".tar":    true,
+	".tar.gz": true,
+	".tar.xz": true,
```
vaikas, line 24: wonder if should also add .tgz here as well?

pkg/action/programkind.go
```diff
@@ -125,3 +134,21 @@ func byExtension(path string) (bool, string) {
 	ret, ok := extMap[filepath.Ext(path)]
 	return ok, ret
 }
+
+// getExt returns the extension of a file path
+// and attempts to avoid including fragments of filenames with other dots before the extension.
+func getExt(path string) string {
```
vaikas, line 140: I'm curious if it would be more resistant to start with the lastIndex in case the actual file name contains `.` characters and we're only caring about the last potentially two.
First lastIndex -> returns the very last,
Second lastIndex -> returns the second last, as long as you feed it everything before the one from above.

vaikas: Also, could we add some tests for parsing these 🙏

pkg/action/archive.go
egibs: Added in [`424a907` (#144)](https://github.com/chainguard-dev/bincapz/pull/144/commits/424a907f4bbc7201f6f4bc242c2e2b0ac06ec853).

pkg/action/archive.go
egibs: Updated in [`424a907` (#144)](https://github.com/chainguard-dev/bincapz/pull/144/commits/424a907f4bbc7201f6f4bc242c2e2b0ac06ec853).

pkg/action/programkind.go
egibs: Added in [`424a907` (#144)](https://github.com/chainguard-dev/bincapz/pull/144/commits/424a907f4bbc7201f6f4bc242c2e2b0ac06ec853).

pkg/action/programkind.go
egibs: Updated the `getExt` function and added tests in [`424a907` (#144)](https://github.com/chainguard-dev/bincapz/pull/144/commits/424a907f4bbc7201f6f4bc242c2e2b0ac06ec853).

Tests:
```sh
❯ go clean -testcache && make test
go test ./... -v
...
=== RUN   TestGetExt
=== RUN   TestGetExt/testdata/file_no_ext
=== RUN   TestGetExt/testdata/file.apk
=== RUN   TestGetExt/testdata/file.jar
=== RUN   TestGetExt/testdata/file.tar
=== RUN   TestGetExt/testdata/file.tgz
=== RUN   TestGetExt/testdata/file.tar.gz
=== RUN   TestGetExt/testdata/file.tar.xz
=== RUN   TestGetExt/testdata/file.zip
=== RUN   TestGetExt/testdata/file_1.0.0.zip
=== RUN   TestGetExt/testdata/file_1.0.0.tar.gz
=== RUN   TestGetExt/testdata/file_1.0.0.tar.xz
=== RUN   TestGetExt/testdata/file_1.0.0.tar
=== RUN   TestGetExt/testdata/file_1.0.0.tgz
=== RUN   TestGetExt/testdata/file_1.0.0.apk
=== RUN   TestGetExt/testdata/file_1.0.0.jar
=== RUN   TestGetExt/testdata/file_1.0.0
=== RUN   TestGetExt/testdata/file.a.b.c.tar.gz
=== RUN   TestGetExt/testdata/file_a.b.c.tar.xz
=== RUN   TestGetExt/testdata/file_a.b.0.tar
--- PASS: TestGetExt (0.00s)
    --- PASS: TestGetExt/testdata/file_no_ext (0.00s)
    --- PASS: TestGetExt/testdata/file.apk (0.00s)
    --- PASS: TestGetExt/testdata/file.jar (0.00s)
    --- PASS: TestGetExt/testdata/file.tar (0.00s)
    --- PASS: TestGetExt/testdata/file.tgz (0.00s)
    --- PASS: TestGetExt/testdata/file.tar.gz (0.00s)
    --- PASS: TestGetExt/testdata/file.tar.xz (0.00s)
    --- PASS: TestGetExt/testdata/file.zip (0.00s)
    --- PASS: TestGetExt/testdata/file_1.0.0.zip (0.00s)
    --- PASS: TestGetExt/testdata/file_1.0.0.tar.gz (0.00s)
    --- PASS: TestGetExt/testdata/file_1.0.0.tar.xz (0.00s)
    --- PASS: TestGetExt/testdata/file_1.0.0.tar (0.00s)
    --- PASS: TestGetExt/testdata/file_1.0.0.tgz (0.00s)
    --- PASS: TestGetExt/testdata/file_1.0.0.apk (0.00s)
    --- PASS: TestGetExt/testdata/file_1.0.0.jar (0.00s)
    --- PASS: TestGetExt/testdata/file_1.0.0 (0.00s)
    --- PASS: TestGetExt/testdata/file.a.b.c.tar.gz (0.00s)
    --- PASS: TestGetExt/testdata/file_a.b.c.tar.xz (0.00s)
    --- PASS: TestGetExt/testdata/file_a.b.0.tar (0.00s)
PASS
...
```

pkg/action/archive.go
```diff
@@ -33,25 +36,27 @@ func copyArchive(src string, dst string) error {
 
 	defer func() {
 		if cerr := w.Close(); cerr != nil {
-			log.Printf("failed to close file: %v", cerr)
+			err = fmt.Errorf("failed to close file: %v", cerr)
```
vaikas, line 39: love the idea, and actually was thinking about proposing this earlier, but there's plenty of other places where we don't actually deal with errors, so didn't want to make you jump through the hoops.

I think (and I could be totally wrong here...) You would have to make the return `error` be a named return argument, because this defer will run after the ones below and will have no effect on setting it. So, I think to make this return in case of close failure, we'd have to:
 * change the signature to return (err error)
 * but then the problem is that if the L44 returns an error, this defer might fail and if it does it will overwrite the error
 * so maybe use something like errors.Join (which will ignore nil values, so I think it should "just work").

Here's one article I found:
https://riandyrn.medium.com/golang-magic-modify-return-value-using-deferred-function-ed0eabdaa75

After having said all though, since we don't check the error on r.Close on L30, I think just logging this as an error here will be fine.

vaikas (commented):
Thanks so much! Love the tests ❤️

pkg/action/archive.go
egibs: Makes sense! I'll just log the error for now.

egibs: Changed in [`be839e6` (#144)](https://github.com/chainguard-dev/bincapz/pull/144/commits/be839e6d7ef7147bc93db2321cf097e64f2c4e5f).

vaikas (approved):
Thank you!

Issue #99, opened 2024-04-03
Is pkg-config a dependency?


alcidesmig:
[alcides@mig] ~ $ GO111MODULE=on go install github.com/chainguard-dev/bincapz@latest
github.com/hillu/go-yara/v4: exec: "pkg-config": executable file not found in $PATH

System:

OS: Arch Linux x86_64
Kernel: 6.6.23-1-lts
yara 4.5.0-1
go version go1.22.1 linux/amd64

tstromberg:
I didn't realize it, but yes, `pkg-config` appears to be an indirect dependency for most users according to https://github.com/hillu/go-yara

> On Unix-like systems, libyara version 4.3, corresponding header files, and pkg-config must be installed. Adding go-yara v4 to a project with Go Modules enabled, simply add the proper dependency…

tstromberg:
Closing, as the answer is, unfortunately, yes.

PR #145, opened 2024-04-17
Mention pkg-config dependency and container image

tstromberg:
Related to #99


README.md
```diff
@@ -24,15 +24,28 @@ Enumerates program capabilities and malicious behaviors using fragment analysis.
 
 ## Requirements
 
-* go 1.21+
-* yara 4.3+ library - you can use this one-liner to install it if necessary:
+A container runtime environment such as Podman or Docker, or local developer tools:
+
+* [go](https://go.dev/) 1.21+
+* [pkg-config](https://www.freedesktop.org/wiki/Software/pkg-config/) - included in many UNIX distributions
+* [yara](https://virustotal.github.io/yara/) - you can use this one-liner to install it:
```
vaikas, line 31: The one liner comment is now in a wrong place?

And maybe rejigger this to now live in the 'Local' section below? Since they are not needed for the containerized version?

tstromberg: good spot - PTAL.

vaikas (approved):

PR #150, opened 2024-04-19
Tune packer rules to avoid false-positives

tstromberg:
I was seeing false-positives with these new rules in Wolfi, particularly with bazel. They are a bit more cautious now.

This now makes use of rule inheritance.


rules/evasion/packer/high_entropy.yara
```diff
@@ -1,9 +1,24 @@
 import "math"
 
-rule high_entropy_7 : suspicious {
+rule smallBinary {
+	condition:
+		// matches ELF or machO binary
```
vaikas, line 5: ❤️

rules/evasion/packer/zstd.yara
```diff
@@ -7,9 +7,9 @@ rule custom_zstd : suspicious {
     strings:
 		// Thanks to https://github.com/VaccinatorSec/yara-rules/blob/master/rules/compressed.yar
 		$ref = {28 B5 2F FD}
-		$ref2 = "zstd"
+		$ref2 = "ZSTD_decompressStream"
     condition:
 		// matches elf or macho
-	    filesize < 52428800 and (uint32(0) == 1179403647 or uint32(0) == 4277009102 or uint32(0) == 3472551422 or uint32(0) == 4277009103 or uint32(0) == 3489328638 or uint32(0) == 3405691582 or uint32(0) == 3199925962) and $ref and $ref2 and math.entropy(1200,filesize) > 6
+	    filesize < 64MB and (uint32(0) == 1179403647 or uint32(0) == 4277009102 or uint32(0) == 3472551422 or uint32(0) == 4277009103 or uint32(0) == 3489328638 or uint32(0) == 3405691582 or uint32(0) == 3199925962) and $ref and $ref2 and math.entropy(1200,filesize) > 7.5
```
vaikas, line 13: And the reason why this can't be:
```
smallBinary and $ref and $ref2 and math.entropy(1200,filesize) > 7.5
```

Is because it breaks yara command line?

vaikas (approved):

rules/evasion/packer/zstd.yara
tstromberg: yeah; cross-file dependencies are generally frowned upon in the YARA world because it breaks rule portability. I haven't tested it, but it's possible it would break bincapz as well. :/

PR #154, opened 2024-04-20
markdown: Include evidence column and URL links

tstromberg:
This adds the "evidence" column back to markdown output, and provides URL links for additional context:

* The key field now contains a RuleURL that points to the original YARA rule
* The description field now contains a ReferenceURL that points to more information about a capability.
* The author credit now has a URL 
* The evidence column is back! It also contains a link, typically to Github searches.


tstromberg:
I'm bringing this back in the form of smaller PR's.

PR #169, opened 2024-04-26
Improve rules based on LightSpy + add Huntress to third_party

tstromberg:
I tested bincapz against the samples noted in https://www.huntress.com/blog/lightspy-malware-variant-targeting-macos and made some improvements.

This does pull in a handful of YARA rules that huntress provides, and includes a Makefile rule to update them in the future. MIT Licensed.


Makefile
```diff
@@ -56,3 +56,17 @@ update-yaraforge:
 	mkdir -p out
 	curl -sL -o out/yaraforge.zip https://github.com/YARAHQ/yara-forge/releases/latest/download/yara-forge-rules-full.zip
 	unzip -o -j out/yaraforge.zip packages/full/yara-rules-full.yar -d rules/third_party/
+
+
+.PHONY: update-huntress
+update-huntress:
+	mkdir -p out
+	curl -sL -o out/huntress.zip https://github.com/huntresslabs/threat-intel/archive/refs/heads/main.zip
```
vaikas, line 64: There doesn't seem to be any releases or tags and I can't figure out from README how you got to this url. Reason I started to look into it, is there no checksumming here available to validate the download?

tstromberg: We can use git, but hardcoding a checksum here would be against the rule's intent: to grab the latest version of something so that we can pin it locally.

Hardcoding a checksum here would be a maintenance burden as it's a double pinning.

vaikas: How do you mean double-pinning? Just curling anything with no checks just seems bad in general? Even having a comment, or a link to where this is generated would be helpful. Like I said, just looking at README, I couldn't figure out where this URL came from.

PR #170, opened 2024-04-26
Upgrade to go1.22 and add goreleaser job checker

cpanato:


tstromberg:
@cpanato - I noticed this error in check_releaser:

` ⨯ release failed after 2s                  error=failed to build for darwin_amd64_v1: exit status 1: error obtaining VCS status: exit status 128
	Use -buildvcs=false to disable VCS stamping.`

Do you still want to merge this?

cpanato:
> @cpanato - I noticed this error in check_releaser:
> 
> ` ⨯ release failed after 2s error=failed to build for darwin_amd64_v1: exit status 1: error obtaining VCS status: exit status 128 Use -buildvcs=false to disable VCS stamping.`
> 
> Do you still want to merge this?

we will merge if i can have a green check, otherwise no, but i am struggling in having a cross compiling for this, we might want to build in each runner platform, like, build mac on macos runner....

but for that and if we use goreleaser we will need a license, or give up on goreleaser and run in another way,

tstromberg:
Ugh. I think the best thing to do is drop the binary release process for now.

After giving it some thought, I think binary images aren't going to be helpful for users - as few of them will have the right version of the dependency installed (libyara). I'd rather users just use the image or compile bincapz.

PR #177, opened 2024-05-01
Mask Chrome extension IDs in threat-hunting keyword list

### LINKED ISSUE #176: XProtect flags bincapz as malware
This is at bincapz HEAD with Sonoma 14.4.1 and XProtect signature base 2192 

Screenshot 2024-05-01 at 8 44 31 AM

tstromberg:
Fixes #176 

XProtect looks for programs that match 3 or more Chrome extension IDs that are listed by the 3rd party threat-hunting YARA set. This obfuscates the listed extension ID slightly by moving the last character to sit within brackets.


egibs:
Thanks for figuring this out!

vaikas (approved):

tstromberg:
@egibs - The idea is that the `make update-XXX` rule should fully automate everything relating to rule updates. With this update, it still will - even with the double pinning.

Eventually, I'd like to have the update rules run automated by a periodic GitHub action. I want to flesh out a few more examples first though.

PR #179, opened 2024-05-02
remove release job, it does not work as is and need more things

cpanato:
related to https://github.com/chainguard-dev/bincapz/pull/170

binary releases are currently difficult today due to the libyara C library dependency, specifically:

- cross-compiling doesn't work w/ go-releaser
- users are required to have the exact libyara version we compiled against to avoid crashes

For now, let's just direct end-users build the software themselves or use the image.


tstromberg (approved):

egibs:
Bummer. I can spend some cycles on figuring this out.

cpanato:
we need to build in a dedicated runner for now to avoid cross-compilation, which will be a bit harder, but if we use Goreleaser, we will need a paid license to use the split build option. i will figure out something

PR #182, opened 2024-05-04
Update to Go 1.22

egibs:
This PR updates from Go `1.21.6` to `1.22` and also contains `go get -u` and `go mod tidy` changes.

This PR also uses `go.mod` for the `go-version-file` to make the CI setup steps easier to manage (rather than specifying `1.22` for each of the `setup-go` versions).


tstromberg:
Just curious - was there a particular inspiration for this change? 

I don't mind upgrading to the latest go version to use a feature, but upping the installation requirements otherwise might exclude some users.

egibs:
> Just curious - was there a particular inspiration for this change?
> 
> I don't mind upgrading to the latest go version to use a feature, but upping the installation requirements otherwise might exclude some users.

Good point. This isn't a necessity right now and it can always be re-opened later. I'll probably port the `go-version-file` changes to another PR, though.

PR #186, opened 2024-05-05
Use go.mod for setup-go; update golangci-lint

egibs:
CI-focused version of #182 that uses `go.mod` for the `setup-go` steps and bumps `golangci-lint` to the latest version while removing some deprecated config (`linters-settings.staticcheck` and `linters-settings.unused`).


tstromberg (approved):

tstromberg:
Thank you!

PR #196, opened 2024-05-08
Disable poorly performing rules, including those with warnings

tstromberg:
Old behavior:
- Rules marked as bad were still executed. These results were used in the file risk calculation even if we later hid the matching rule result (see "ls.mdiff" test result)
- Rules with warnings were still executed, with the warning hidden unless `-verbose` was passed.

New behavior:
- Rules marked as bad are disabled
- Rules with warnings are:
  - Disabled unless there is an exception
  - If an unexpected rule has a warning, log an error

Other notes:
- I've removed the e-mail rule, as it had a warning and wasn't useful
- I brought JSON tests back from the dead to prove the "ls" data. We had purged these tests a while back when the JSON output wasn't yet stable.
- This PR does not provide a performance improvement or regression, but it does fix a bug!

### DIFF
--- a/pkg/compile/compile.go
+++ b/pkg/compile/compile.go
@@ -17,6 +17,23 @@ import (
 
 var FS = rules.FS
 
+// badRules are noisy 3rd party rules to silently disable.
+var badRules = map[string]bool{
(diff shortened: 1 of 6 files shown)


egibs (approved):
LGTM!

PR #200, opened 2024-05-08
make archive extraction problems non-fatal

tstromberg:
I wanted to capture `--stats` for https://github.com/lxyeternal/pypi_malregistry - but it contained an archive that wasn't valid. 

Now it outputs an error to stderr:

`time=2024-05-08T12:08:27.563-04:00 level=ERROR msg="unable to process /Users/t/src/malware/pypi_malregistry/distrib/0.1/distrib-0.1.tar.gz: extract to temp: extract: extract tar: failed to create gzip reader: gzip: invalid header"`

While trying to unwind where the error message was, I saw there was some opportunity to simplify the code, so I did so.

### DIFF
--- a/pkg/action/archive.go
+++ b/pkg/action/archive.go
@@ -9,7 +9,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"path"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


egibs (approved):
Appreciate the simplification!

Issue #204, opened 2024-05-08
bincapz OOM'd at 8GB of RAM with a large source tree


tstromberg:
bincapz consumed. >8GB of RAM using it against a large directory:

```
[24669.706944] Out of memory: Killed process 61135 (bincapz) total-vm:8580228kB, anon-rss:2136184kB, file-rss:0kB, shmem-rss:92kB, UID:1000 pgtables:13940kB oom_score_adj:0
[24676.379361] oom_reaper: reaped process 61135 (bincapz), now anon-rss:32kB, file-rss:0kB, shmem-rss:92kB
```

bincapz worked on this directory a few weeks ago, though a lot has changed since then.

egibs:
I added some profiling and ran `bincapz` against my local Downloads directory to get a decently-long run:
```
(pprof) sample_index = alloc_space
(pprof) top10
Showing nodes accounting for 2.96GB, 85.87% of 3.44GB total
Dropped 169 nodes (cum <= 0.02GB)
Showing top 10 nodes out of 108
      flat  flat%   sum%        cum   cum%
    1.48GB 43.14% 43.14%     1.74GB 50.44%  io.copyBuffer
    0.60GB 17.45% 60.59%     0.60GB 17.45%  github.com/liamg/magic.worker
    0.27GB  7.70% 68.30%     0.40GB 11.72%  github.com/hillu/go-yara/v4.(*Rule).getMatchStrings
    0.15GB  4.31% 72.61%     0.15GB  4.31%  compress/flate.(*huffmanDecoder).init
    0.14GB  3.93% 76.54%     0.14GB  3.93%  github.com/rivo/uniseg.NewGraphemes
    0.11GB  3.19% 79.73%     0.11GB  3.19%  github.com/hillu/go-yara/v4.(*String).Matches
    0.10GB  2.98% 82.71%     0.10GB  2.98%  github.com/ulikunitz/xz/lzma.(*decoder).readOp
    0.05GB  1.54% 84.25%     0.07GB  1.93%  github.com/chainguard-dev/bincapz/pkg/report.matchStrings
    0.03GB  0.81% 85.06%     0.06GB  1.60%  github.com/chainguard-dev/bincapz/pkg/compile.Recursive.func1
    0.03GB  0.81% 85.87%     0.08GB  2.44%  github.com/chainguard-dev/bincapz/pkg/action.programKind
```

[...]

PR #207, opened 2024-05-10
Add profiling to help with OOM/performance investigations

egibs:
This is a quick PR to help with identifying the cause(s) of #204 which adds CPU and Memory profiling along with tracing.

If anything, I wanted the code to be available to aid with the investigation -- this doesn't necessarily need to be merged; however, as the number of capabilities grows it would be nice to have some insight into any causes of performance regressions.

### DIFF
--- a/bincapz.go
+++ b/bincapz.go
@@ -9,6 +9,7 @@ import (
 	"flag"
 	"fmt"
 	"io/fs"
+	"log"
 	"log/slog"
 	"os"
 	"strings"
@@ -16,6 +17,7 @@ import (
 	"github.com/chainguard-dev/bincapz/pkg/action"
 	"github.com/chainguard-dev/bincapz/pkg/bincapz"
 	"github.com/chainguard-dev/bincapz/pkg/compile"
+	"github.com/chainguard-dev/bincapz/pkg/profile"
 	"github.com/chainguard-dev/bincapz/pkg/render"
 	"github.com/chainguard-dev/bincapz/rules"
 	thirdparty "github.com/chainguard-dev/bincapz/third_party"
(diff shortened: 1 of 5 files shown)


bincapz.go
```diff
@@ -9,13 +9,15 @@ import (
 	"flag"
 	"fmt"
 	"io/fs"
+	"log"
```
tstromberg, line 12: Rather than adding a second logging library, please re-use the `clog` import - even if, admittedly, I prefer the built-in Go library.

pkg/profile/profile.go
```diff
@@ -0,0 +1,87 @@
...
+		trace.Stop()
+		c.Close()
+		m.Close()
+		t.Close()
+	}
+
+	return stop, nil
+}
+
+func HandleProfileFlag(pf *bool) error {
```
tstromberg, line 77: Please move all flag handling to `main`.

tstromberg (changes requested):
This is great! I'm going to use this today.

bincapz.go
egibs: Addressed in [`c978170` (#207)](https://github.com/chainguard-dev/bincapz/pull/207/commits/c978170042c5bc28de6d7c8a5739f7b2e4095e22).

pkg/profile/profile.go
egibs: Addressed in [`c978170` (#207)](https://github.com/chainguard-dev/bincapz/pull/207/commits/c978170042c5bc28de6d7c8a5739f7b2e4095e22).

bincapz.go
```diff
@@ -111,6 +122,10 @@ func main() {
 
 	err = renderer.Full(ctx, *res)
 	if err != nil {
-		clog.Fatal("render failed", slog.Any("error", err))
+		log.Fatal("render failed", slog.Any("error", err))
+	}
+
+	if *profileFlag {
```
egibs, line 128: This avoids `os.Exit(1)` and `defer` not playing nicely together.

tstromberg (approved):
nice work!

PR #210, opened 2024-05-10
Add version support

egibs:
Closes: https://github.com/chainguard-dev/bincapz/issues/209

This PR adds support for a `--version` flag that will either use the newly-added `VERSION` file or check for the `main.version` build key.

Testing:
```
❯ go run . --version
bincapz 0.11.0
❯ go build -ldflags "-X main.version=$(cat VERSION)" && ./bincapz --version
bincapz 0.11.0
```

We aren't publishing binaries yet but this should work the same if we use the `VERSION` file as the source of truth; however, this file will need to be updated prior to any new releases being cut. 

If there's a better way to handle this I'm all ears!


VERSION
```diff
@@ -0,0 +1 @@
+0.11.0
```
tstromberg, line 1: I think we can get rid of this now.

bincapz.go
```diff
@@ -50,6 +51,8 @@ func main() {
 	statsFlag := flag.Bool("stats", false, "Show statistics about the scan")
 	thirdPartyFlag := flag.Bool("third-party", true, "Include third-party rules, which may have licensing restrictions")
 	verboseFlag := flag.Bool("verbose", false, "Emit verbose logging messages to stderr")
+  versionFlag := flag.Bool("version", false, "show version information")
```
tstromberg: Capitalize "Show" for consistency with your more recent flag changes.

egibs: Fixed in [`9b582f9` (#210)](https://github.com/chainguard-dev/bincapz/pull/210/commits/9b582f9874505df5023c7beff8bdb9a24e993778).

VERSION
egibs: Removed in [`5be38e1` (#210)](https://github.com/chainguard-dev/bincapz/pull/210/commits/5be38e1648fa83c7a9330a33a247efd1461a173e).

tstromberg (approved):

PR #214, opened 2024-05-10
Remove unnecessary archive code; fix .tar.xz support

egibs:
I was running a scan on the `go1.22.2.linux-arm64.tar.gz` archive which contains nested `.tar` and `.zip` files. 

The recursive extraction logic in each of the `extract` functions was indeed incorrect, not needed, and broke extraction for this type of archive.

Removing this logic 1) fixed the issue with that archive and 2) was proven unnecessary for `.gem` files:
```
❯ ls -hal /var/folders/3g/88131l9j11x995ppjbxsvhbh0000gn/T/cocoapods-fixbugs-plugin-0.1.0.gem2017939464
Permissions Size User  Date Modified Name
.rw-r--r--@   76 egibs 10 May 16:27  .travis.yml
drwxr-xr-x@    - egibs 10 May 16:27  bin
.rw-r--r--@  418 egibs 10 May 16:27  checksums.yaml
.rw-r--r--@  991 egibs 10 May 16:27  cocoapods-fixbugs-plugin.gemspec
.rw-r--r--@  109 egibs 10 May 16:27  Gemfile
drwxr-xr-x@    - egibs 10 May 16:27  lib
.rw-r--r--@ 1.1k egibs 10 May 16:27  LICENSE.txt
.rw-r--r--@ 2.2k egibs 10 May 16:27  metadata
.rw-r--r--@   28 egibs 10 May 16:27  Rakefile
.rw-r--r--@ 1.4k egibs 10 May 16:27  README.md
```

### DIFF
--- a/pkg/action/archive.go
+++ b/pkg/action/archive.go
@@ -44,7 +44,8 @@ func extractTar(ctx context.Context, d string, f string) error {
 		}
 		defer gzStream.Close()
 		tr = tar.NewReader(gzStream)
-	} else if strings.Contains(filename, ".tar.xz") {
+	}
+	if strings.Contains(filename, ".tar.xz") {
 		xzStream, err := xz.NewReader(tf)
 		if err != nil {
 			return fmt.Errorf("failed to create xz reader: %w", err)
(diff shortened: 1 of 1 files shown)


pkg/action/archive.go
```diff
@@ -300,8 +263,12 @@ func extractArchiveToTempDir(ctx context.Context, path string) (string, error) {
 	}
 
 	for file := range extractedFiles {
-		if err := extractNestedArchive(ctx, tmpDir, file, extractedFiles); err != nil {
-			return "", fmt.Errorf("extract nested archive: %w", err)
+		ext := getExt(file)
+		if _, ok := archiveMap[ext]; ok {
+			fmt.Printf("found a nested archive %s", file)
+			if err := extractNestedArchive(ctx, tmpDir, file, extractedFiles); err != nil {
```
egibs, line 269: It made way more sense to only focus on one archive file at a time within `extractNestedArchive` instead of having it handle the iteration.

tstromberg (approved):
simpler is good!

once we are able to stabilize the output of an archive scan, we're going to want tests!

PR #220, opened 2024-05-13
Add -o flag to set output destination

tstromberg:
This way callers don't need to perform stdout buffer management.


bincapz.go
```diff
@@ -80,7 +80,19 @@ func main() {
...
-	renderer, err := render.New(*formatFlag, os.Stdout)
+	outFile := os.Stdout
+	var err error
+	if *outputFlag != "" {
+		outFile, err = os.OpenFile(*outputFlag, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0o600)
+		if err != nil {
+			log.Fatal("open file", slog.Any("error", err), slog.String("path", *outputFlag))
+		}
+	}
+	defer func() {
```
egibs, line 91: Given:
```
exitAfterDefer: log.Fatal will exit, and `defer func(){...}(...)` will not run (gocritic)
```
would it make sense to do `if err := outFile.Close(); err != nil { ... }` here?

tstromberg: I've re-arranged things so that defer works - it also adds some initial exit code handling too!

egibs (approved):

PR #223, opened 2024-05-13
version bump to v0.12.0

tstromberg:
This does change it so that versions are prefixed with "v" for clarity. Please let me know what you think!


egibs (commented):
Works for me; the leading `v` is probably the most correct way to tag a release.

Also, can you update `bincapz.go` to use the new return codes? I pushed that commit too late to include it in #210.

```go
if *versionFlag {
		ver, err := version.Version()
		if err != nil {
			fmt.Printf("bincapz unknown version\n")
		}
		fmt.Printf("%s\n", ver)
		returnCode = ExitOK
		return
	}
```

egibs (approved):

PR #228, opened 2024-05-17
Replace Behavior map with slice

egibs:
Closes: https://github.com/chainguard-dev/bincapz/issues/208

This PR replaces the original `map[string]*Behavior` with `[]*Behavior`. 

Moving from maps to slices complicates things a bit but mainly around not being able to rely on the `key`. To work around this, I added a new `Evidence` field to the `Behavior` struct to store the key which handles all of the previous key operations.

And the good news is: `make test` runs cleanly with these changes:
```
PASS
ok  	github.com/chainguard-dev/bincapz/samples	16.336s
```

### DIFF
--- a/pkg/action/diff.go
+++ b/pkg/action/diff.go
@@ -76,7 +76,7 @@ func Diff(ctx context.Context, c Config) (*bincapz.Report, error) {
 
 		rbs := &bincapz.FileReport{
 			Path:              tr.Path,
-			Behaviors:         map[string]*bincapz.Behavior{},
+			Behaviors:         []*bincapz.Behavior{},
 			PreviousRiskScore: fr.RiskScore,
 			PreviousRiskLevel: fr.RiskLevel,
 			RiskLevel:         tr.RiskLevel,
@@ -84,10 +84,17 @@ func Diff(ctx context.Context, c Config) (*bincapz.Report, error) {
 		}
 
 		// if source behavior is not in the destination
-		for key, b := range fr.Behaviors {
-			if _, exists := tr.Behaviors[key]; !exists {
-				b.DiffRemoved = true
-				rbs.Behaviors[key] = b
+		for _, fb := range fr.Behaviors {
+			found := false
+			for _, tb := range tr.Behaviors {
+				if tb.Evidence == fb.Evidence {
+					found = true
+					break
+				}
+			}
+			if !found {
+				fb.DiffRemoved = true
+				rbs.Behaviors = append(rbs.Behaviors, fb)
 			}
 		}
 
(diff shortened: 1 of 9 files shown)


pkg/report/report.go
```diff
@@ -311,9 +311,10 @@ func Generate(ctx context.Context, path string, mrs yara.MatchRules, ignoreTags
 		ruleURL := generateRuleURL(m.Namespace, m.Rule)
 
 		b := &bincapz.Behavior{
-			RiskScore:    risk,
-			RiskLevel:    RiskLevels[risk],
+			Key:          key,
```
tstromberg, line 314: nit: Can you rename this field ID?

IMHO, it's more important than just being a key; it's a unique ID for this behavwior.

pkg/bincapz/bincapz.go
```diff
@@ -21,20 +21,23 @@ type Behavior struct {
 
 	DiffAdded   bool `json:",omitempty" yaml:",omitempty"`
 	DiffRemoved bool `json:",omitempty" yaml:",omitempty"`
+
+	// Key is the original map key from map[string]*Behavior
```
tstromberg, line 25: Update this description as well, to something like:

// ID is a unique behavior identifier

pkg/action/diff.go
```diff
@@ -61,39 +61,87 @@ func Diff(ctx context.Context, c Config) (*bincapz.Report, error) {
...
+		Path:              tr.Path,
+		Behaviors:         []*bincapz.Behavior{},
+		PreviousRiskScore: fr.RiskScore,
+		PreviousRiskLevel: fr.RiskLevel,
+		RiskLevel:         tr.RiskLevel,
+		RiskScore:         tr.RiskScore,
 	}
+}
 
+func behaviorExists(
```
tstromberg, line 127: For consistency, please keep function definitions to a single line.

This also encourages smaller variable names and encourages people to not introduce functions with an excessive number (>4) of positional arguments. We should add a lint check for that :)

tstromberg (changes requested):
This looks great - I can't imagine how much trial and error it took to get this PR right given how much we relied on map behavior.

pkg/report/report.go
egibs: Addressed in [`290f13e` (#228)](https://github.com/chainguard-dev/bincapz/pull/228/commits/290f13e54bcde6059760f1807cfb98be377f024c).

pkg/action/diff.go
egibs: Makes sense -- some of these new functions have >4 arguments but going forward I'll keep that in mind.

tstromberg (approved):

PR #236, opened 2024-05-26
Miscellaneous tweaks, additions, and performance improvements

egibs:
This PR has a few tweaks to consolidate string matching logic and improve performance along with some other miscellaneous changes.

On the performance side, I replaced `filepath.Walk` with `filepath.WalkDir` when recursively looking for files in a given scan path. With the code in `main`, scanning the Linux repository consisting of ~85,000 files took roughly ~4 minutes. With these changes, the scan takes about ~3 1/2 which is a decent improvement.

The other changes are mostly just syntax changes, though I did add a few more program kinds for the sake of completion along with a test and benchmark for `longestUnique` since it looks like a possible area for future optimization.


Makefile
```diff
@@ -49,7 +49,11 @@ fix: $(FIXERS)
 
 .PHONY: test
 test:
-	go test ./... -v
+	go test ./...
```
egibs, line 52: `-v` made it hard to track down failures since we do a lot of debug logging.

tstromberg (approved):
Looks like a great set of improvements all around. Thank you!

PR #239, opened 2024-05-28
Remove leading @ from Author value

tstromberg:
This will prevent people from accidentally pinging authors every time bincapz output is pasted into a GitHub PR or Issue.


egibs (approved):
Good catch!

PR #240, opened 2024-06-02
Add Kiteshield rule

egibs:
Closes: https://github.com/chainguard-dev/bincapz/issues/238

This PR implements the rule mentioned in the `xlab` [blog post](https://blog.xlab.qianxin.com/kiteshield_packer_is_being_abused_by_linux_cyber_threat_actors/) and also adds each of the IOC MD5 hashes to the Rule metadata.


rules/tools/kiteshield/kiteshield.yara
```diff
@@ -0,0 +1,39 @@
...
+        $loader_s4 = {cf c0 da d6 d5 cd c5 c5 ca c8}
+        // "LD_AUDIT"
+        $loader_s5 = {cf c0 da c7 d2 cc c0 de}
+        // "LD_DEBUG"
+        $loader_s6 = {cf c0 da c2 c2 ca dc cd}
+        // "0123456789abcdef"
+        $loader_s7 = {b3 b5 b7 b5 b3 bd bf bd b3 b5 ec ec ec f4 f4 f4}
+
+    condition:
+	    $loader_jmp and all of ($loader_s*) and elf.type==elf.ET_EXEC and elf.machine == elf.EM_X86_64
```
egibs, line 38: I wasn't sure if doing `any of (elf.machine)` as opposed to `elf.machine == elf.EM_X86_64` would provide any benefit so I left it as-is.

tstromberg (approved):
This is a solid start. Things we may want to change in future PR's are:

- Remove dependency on "elf" import (it's surprisingly slow)
- Add arm64 support
-

PR #246, opened 2024-06-05
Makefile: add 'refresh-sample-testdata' rule

tstromberg:
Updating sample testdata can be confusing for new contributors. This should make it easier.

NOTE: Occasionally, this may generate an unexpected whitespace diff for markdown samples. Our markdown tests are whitespace insensitive.


egibs (commented):
Doing something like this crossed my mind a couple of weeks ago -- this is much better than looking back in my history for the correct commands.

One thought: should the `test` target call this target first so that the test data is always valid?

egibs (approved):
+approving since my question is non-blocking

tstromberg:
> One thought: should the `test` target call this target first so that the test data is always valid?

It sounds like a good idea, but in my experience, any time something named "test" modifies data, developers find it to be an unexpected and baffling behavior.

egibs:
> It sounds like a good idea, but in my experience, any time something named "test" modifies data, developers find it to be an unexpected and baffling behavior.

True indeed. This is always one `make` command away so that's still convenient enough. 🚀

PR #249, opened 2024-06-06
Add new --min-*-risk flags, obsolete --min-level

tstromberg:
The generic "level" and "score" terminology was not self-describing. This changes our flags to specifically mention "risk". It also allows you to pass in a level name like "low" rather than an integer.

* Old style: `--min-level=2`
* New style: `--min-risk=medium`

Fixes https://github.com/chainguard-dev/bincapz/issues/231


bincapz.go
```diff
@@ -35,6 +35,24 @@ var (
 	ExitInvalidArgument = 22
 )
 
+// parse risk levels
```
egibs, line 38:
```suggestion
// parse risk levels.
```

egibs (approved):

PR #253, opened 2024-06-06
Add Workflows to update version and create version tags

egibs:
Closes: https://github.com/chainguard-dev/bincapz/issues/251

This PR adds two Workflows:
- One will create version update PRs via a `workflow_dispatch` (i.e., manually for now)
- One will create tags when commits are pushed to `main` (only if they do not exist)

In practice, the version update PRs will be merged with an updated version string which will then cause the tag Workflow to create a new tag matching that version. 

Every other commit to `main` will result in the tag Workflow exiting without pushing any new tags since the current version's tag will already exist.

### DIFF
--- a/.github/workflows/tag.yaml
+++ b/.github/workflows/tag.yaml
@@ -9,6 +9,9 @@ permissions:
   id-token: write
   contents: write
 
+env:
+  VERSION_FILE: pkg/version/version.go
+
 jobs:
   tag:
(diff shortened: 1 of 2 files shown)


tstromberg (approved):
Thank you!

egibs:
I plan on testing the PR creation at least. If that looks good, we can do real patch PR -> merge -> tag test.

PR #256, opened 2024-06-09
Release/Tag Workflow followup fixes

egibs:
This PR makes a few tweaks/additions:
1. Uses `sed -i` instead of `sed -i''` (MacOS testing artifact)
2. Adds version output to the version bump Workflow for quick verification; uses a branch variable (I typo'd the branch name inadvertently which broke the push)
3. Uses better (and correct) variables for the tag Workflow
4. Uses GPG signing for tags (which will be needed for `gitsign`)
5. Pushes the newly-created tag rather than all tags

Original errors:
CleanShot 2024-06-09 at 17 30 57@2x
CleanShot 2024-06-09 at 17 31 19@2x

### DIFF
--- a/.github/workflows/tag.yaml
+++ b/.github/workflows/tag.yaml
@@ -24,18 +24,20 @@ jobs:
     - name: Get Version
       id: get-version
       run: |
-        CURRENT_VERSION=$(awk -F'"' '/ID string =/ {print $2}' ${{ env.VERSION_FILE }})
-        echo "VERSION=$CURRENT_VERSION" >> "$GITHUB_OUTPUT"
+        VERSION=$(awk -F'"' '/ID string =/ {print $2}' ${{ env.VERSION_FILE }})
+        echo "VERSION=$VERSION" >> "$GITHUB_OUTPUT"
     - name: Create Tag
       if: ${{ steps.get-version.outputs.VERSION != '' }}
       run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/tag.yaml
```diff
@@ -24,18 +24,20 @@ jobs:
...
       if: ${{ steps.get-version.outputs.VERSION != '' }}
       run: |
-        if [ $(git tag -l "${{ steps.get-version.outputs.VERSION }}") ]; then
-            echo "Tag already exists for version $CURRENT_VERSION"
+        VERSION=${{ steps.get-version.outputs.VERSION }}
+        if [ $(git tag -l "$VERSION") ]; then
+            echo "Tag already exists for version $VERSION"
             exit 0
         else
+          git config --global tag.gpgsign true
```
egibs, line 37: This is mentioned in the default Gitsign configuration: https://github.com/sigstore/gitsign?tab=readme-ov-file#configuration

However, the Action does not handle this by default:
https://github.com/chainguard-dev/actions/blob/e82b4e5ae10182af72972addcb3fedf7454621c8/setup-gitsign/action.yml#L103-L111

egibs: Opened a PR to add `git config --global tag.gpgsign true` to the Action: https://github.com/chainguard-dev/actions/pull/442

tstromberg (approved):

PR #260, opened 2024-06-10
Create release instead of Tag; other fixes

egibs:
We can automate tag creation via `gh release create` which is a more common practice than specifically creating a tag from `main` and then creating a release after the fact.

This PR replaces the tag creation Workflow with a release creation Workflow that will run via `workflow_dispatch` and create a draft Release that can be reviewed prior to publishing. This will allow us to control the timing of releases but still automate the tedious parts.

This PR also adds cross-step variables for the version bump Workflow so that we can actually access the correct version information (though the core version logic is working):
CleanShot 2024-06-10 at 07 13 22@2x
and also updates the `certificate-identity-regexp` to include a trailing `/*` so we match subjects appropriately.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,9 +1,7 @@
-name: Tag Release
+name: Cut Release
 
 on:
-  push:
-    branches:
-      - main
+  workflow_dispatch:
 
 permissions:
   id-token: write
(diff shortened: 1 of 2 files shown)


tstromberg (approved):
I'm hoping we can test this today!

PR #265, opened 2024-06-10
Add --err-first-(hit|miss) flags

tstromberg:
This flag is intended to streamline rule optimization for false-positives/false-negatives.

* If you point bincapz at a malware repository, you can now use `--err-first-miss` to error if anything doesn't match.
* If you point bincapz at a known-clean directory, you can now use `--err-first-hit` to error if something matches

This has been tested with archive files and OCI images. The intended behavior is that it should consider the entirety of an archive or image before exiting early for a miss.

After revisiting it for the first time in a couple of months, I found some of the file-scanning code difficult to follow, so I made it more verbose. I also fixed a bug where if the path given was a symlink, bincapz wouldn't follow it.

### DIFF
--- a/pkg/action/scan.go
+++ b/pkg/action/scan.go
@@ -9,6 +9,7 @@ import (
 	"log/slog"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 
 	"github.com/chainguard-dev/bincapz/pkg/bincapz"
@@ -26,6 +27,12 @@ func findFilesRecursively(ctx context.Context, root string, c Config) ([]string,
 	self, err := os.Executable()
 	if err != nil {
 		return nil, fmt.Errorf("abs: %w", err)
+	}
+
+	// Follow symlink if provided at the root
+	root, err = filepath.EvalSymlinks(root)
+	if err != nil {
+		return nil, err
 	}
 
 	err = filepath.WalkDir(root,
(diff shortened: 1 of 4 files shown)


pkg/action/scan.go
```diff
@@ -114,114 +162,167 @@ func recursiveScan(ctx context.Context, c Config) (*bincapz.Report, error) {
...
 
-		for _, p := range rp {
-			isArchive := false
-			ext := getExt(p)
-			if _, ok := archiveMap[ext]; ok {
-				isArchive = true
-			}
-			if isArchive {
-				err = processArchive(ctx, c, yrs, r, p, logger)
+		// path refersll to a real local path, not a virtual scan path
```
egibs, line 190:
```suggestion
		// path refers to a real local path, not a virtual scan path
```

pkg/action/scan.go
```diff
@@ -114,114 +162,167 @@ func recursiveScan(ctx context.Context, c Config) (*bincapz.Report, error) {
...
+			return nil, fmt.Errorf("render: %w", err)
 		}
 	}
-	r.Files[path] = fr
-	return nil
+
+	return fr, nil
 }
 
+// Scan is the public API forYARA scanning a data source, applying filters if necessary.
```
egibs, line 325:
```suggestion
// Scan is the public API for YARA scanning a data source, applying filters if necessary.
```

pkg/action/scan.go
```diff
@@ -28,6 +29,12 @@ func findFilesRecursively(ctx context.Context, root string, c Config) ([]string,
 		return nil, fmt.Errorf("abs: %w", err)
 	}
 
+	// Follow symlink if provided at the root
+	root, err = filepath.EvalSymlinks(root)
```
egibs, line 33: 👍🏻

egibs (commented):
Overall LGTM -- this code is much cleaner and readable (shoutout to more descriptive variable names).

egibs (approved):

PR #279, opened 2024-06-22
Fix version check in verson.yaml Workflow

egibs:
We were validating the updated version string against a regex expression that expected the string to be prefixed with `v` when we only update the actual semver.

This PR removes adds the `v` prefix to the `printf` statements in `version.yaml` to fix [this](https://github.com/chainguard-dev/bincapz/actions/runs/9627113069/job/26553832141) failure:
```
Error: VERSION is not a valid semver
```

and removes the `v` from the version string in this line (original code shown):
```sh
sed -i -e "s/ID string = \"v[0-9]+\.[0-9]+\.[0-9]+\"/ID string = \"v$VERSION\"/" ${{ env.VERSION_FILE }}
```

I tested this locally and it fixed the issue --

Fixed code:
```sh
❯ ./test.sh patch
Current bincapz version: v0.13.1
New bincapz version: v0.13.2
```

Original code:
```sh
❯ ./test.sh patch
v0.13.1
Error: 0.13.2 is not a valid semver
```

### DIFF
--- a/.github/workflows/version.yaml
+++ b/.github/workflows/version.yaml
@@ -42,13 +42,13 @@ jobs:
         
         case "$UPDATE_TYPE" in
           major)
-            VERSION=$(printf "%d.0.0" $((${VERSION_PARTS[0]}+1)))
+            VERSION=$(printf "v%d.0.0" $((${VERSION_PARTS[0]}+1)))
             ;;
           minor)
-            VERSION=$(printf "%s.%d.0" ${VERSION_PARTS[0]} $((${VERSION_PARTS[1]}+1)))
+            VERSION=$(printf "v%s.%d.0" ${VERSION_PARTS[0]} $((${VERSION_PARTS[1]}+1)))
             ;;
           patch)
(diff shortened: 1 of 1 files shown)


.github/workflows/version.yaml
```diff
@@ -56,7 +56,7 @@ jobs:
             ;;
         esac
 
-        if [[ ! "${VERSION}" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+        if [[ ! "${VERSION}" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
```
egibs, line 59: The alternative to this would be to update all of the `printf` statements to prefix the `v` but I think simplifying that portion as much as possible is a good idea.

egibs: Actually, we'll need to do that.

egibs: Changed in [`272b63a` (#279)](https://github.com/chainguard-dev/bincapz/pull/279/commits/272b63ab2fbb5a50ff1aec822df023aefa32bc53).

tstromberg (approved):

Issue #292, opened 2024-06-24
Remove non-matching checksums from rules


tstromberg:
As per https://github.com/chainguard-dev/bincapz/runs/26600819412

Non matching file	Rule	YARA file
[f60c1214b5091e6e4e5e7db0c16bf18a062d096c6d69fe1eb3cbd4c50c3a3ed6](https://www.virustotal.com/gui/file/f60c1214b5091e6e4e5e7db0c16bf18a062d096c6d69fe1eb3cbd4c50c3a3ed6)	[login_records](https://github.com/chainguard-dev/bincapz/blob/49ff7c03328f018fb06ef708599571b7b6608b57/rules/admin/logs/current_logins.yara#L2)	rules/admin/logs/current_logins.yara
[ad16989a3ebf0b416681f8db31af098e02eabd25452f8d781383547ead395237](https://www.virustotal.com/gui/file/ad16989a3ebf0b416681f8db31af098e02eabd25452f8d781383547ead395237)	[login_records](https://github.com/chainguard-dev/bincapz/blob/49ff7c03328f018fb06ef708599571b7b6608b57/rules/admin/logs/current_logins.yara#L2)	rules/admin/logs/current_logins.yara
[589dbb3f678511825c310447b6aece312a4471394b3bc40dde6c75623fc108c0](https://www.virustotal.com/gui/file/589dbb3f678511825c310447b6aece312a4471394b3bc40dde6c75623fc108c0)	[executable_calls_archive_tool](https://github.com/chainguard-dev/bincapz/blob/49ff7c03328f018fb06ef708599571b7b6608b57/rules/archives/tar-command.yara#L2)	rules/archives/tar-command.yara
[99296550ab836f29ab7b45f18f1a1cb17a102bb81cad83561f615f3a707887d7](https://www.virustotal.com/gui/file/99296550ab836f29ab7b45f18f1a1cb17a102bb81cad83561f615f3a707887d7)	[executable_calls_archive_tool](https://github.com/chainguard-dev/bincapz/blob/49ff7c03328f018fb06ef708599571b7b6608b57/rules/archives/tar-command.yara#L2)	rules/archives/tar-command.yara
[...]

PR #293, opened 2024-06-24
Remove Kiteshield false negatives

egibs:
Relates to: #292

This PR removes the six MD5 hashes that appear as false negatives in Yara CI.

Even with the original rule logic documented in the reference blog post they show up as false negatives: https://github.com/egibs/bincapz/runs/26625762043

After removing these hashes, there are no more false negatives for the Kiteshield Rule: https://github.com/egibs/bincapz/runs/26625841070

### DIFF
--- a/rules/evasion/packer/kiteshield.yara
+++ b/rules/evasion/packer/kiteshield.yara
@@ -1,21 +1,13 @@
-import "elf"
-
 rule kiteshield : high {
   meta:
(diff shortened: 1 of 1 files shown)


tstromberg (approved):

tstromberg:
It's possible it's a YARA-CI bug with handling MD5 hashes - normally these are SHA256.

PR #300, opened 2024-06-29
Resolve datadog-agent kworker false positives

egibs:
Closes: https://github.com/chainguard-dev/bincapz/issues/288

`"kworker" fullword` was likely matching these strings in `process-agent`:
- `is_kworker`
- `checkworker`

This PR tweaks the `$kworker2` string to match either `kworker` or `Kworker` between word boundaries which will avoid matching against strings like those mentioned above and also adds explicit ignores for `is_kworker` and `checkworker`.

Discovery:
```
$ docker run --rm -it --platform=linux/amd64 --entrypoint sh cgr.dev/chainguard/datadog-agent:latest-dev
...
/ # apk update; apk add vim
/ # xxd -c 64 -g 2 -b /usr/bin/process-agent | grep -i kworker
```

These strings were also present:
- `MarkWorker`
- `IsKworker`
- `IsNotKworker`
- `NotKworker`

The new rule should ignore these as well.


rules/evasion/fake-process-name.yara
```diff
@@ -7,10 +6,11 @@ rule fake_kworker_val : critical {
     hash_2023_Unix_Downloader_Rocke_6107 = "61075056b46d001e2e08f7e5de3fb9bfa2aabf8fb948c41c62666fd4fab1040f"
   strings:
     $kworker = /\[{0,1}kworker\/[\d:\]]{1,5}/
-    $kworker2 = "kworker" fullword
+    $kworker2 = /\b[Kk]worker\b/
     $kworker3 = "[kworker"
+    $ignore_ref = /\w+\_kworker/
```
egibs, line 11:
```suggestion
    $ignore_ref = /\w{0,32}\_kworker/
```

tstromberg (changes requested):

rules/evasion/fake-process-name.yara
```diff
@@ -7,10 +6,11 @@ rule fake_kworker_val : critical {
     hash_2023_Unix_Downloader_Rocke_6107 = "61075056b46d001e2e08f7e5de3fb9bfa2aabf8fb948c41c62666fd4fab1040f"
   strings:
     $kworker = /\[{0,1}kworker\/[\d:\]]{1,5}/
-    $kworker2 = "kworker" fullword
+    $kworker2 = /\b[Kk]worker\b/
```
egibs, line 9:
```suggestion
    $kworker2 = /\bkworker\b/
```

rules/evasion/fake-process-name.yara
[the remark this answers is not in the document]
egibs: Updated in [`6e1ab28` (#300)](https://github.com/chainguard-dev/bincapz/pull/300/commits/6e1ab28c49de5c3ac5de8bbec0e28184d5cda290).

rules/evasion/fake-process-name.yara
```diff
@@ -7,10 +6,14 @@ rule fake_kworker_val : critical {
     hash_2023_Unix_Downloader_Rocke_6107 = "61075056b46d001e2e08f7e5de3fb9bfa2aabf8fb948c41c62666fd4fab1040f"
   strings:
     $kworker = /\[{0,1}kworker\/[\d:\]]{1,5}/
-    $kworker2 = "kworker" fullword
+    $kworker2 = /\bkworker\b/
```
tstromberg, line 9: just for my own education, what's the difference between \b and fullword?

rules/evasion/fake-process-name.yara
```diff
@@ -7,10 +6,14 @@ rule fake_kworker_val : critical {
...
     $kworker3 = "[kworker"
+    // datadog process-agent
+    $ignore_ref = /(data->|[!\*\.])?is_kworker(\.\w{0,32})?/
```
tstromberg, line 12: Can these expressions be simplified? Is there a reason why a single string (non-regex) isn't sufficient to exclude datadog?

tstromberg (changes requested):
The ignore_refs look more complicated than we usually have, so I just wanted to see if they could be made into non-regex form.

Also related: https://github.com/chainguard-dev/bincapz/issues/306

It'd be nice if we had at least one sample to validate that we aren't breaking this query, but we can save that for another PR too.

rules/evasion/fake-process-name.yara
egibs: For the purposes I was using it for, probably not. It would definitely provide more control in a larger expression but I can change this back to `"kworker" fullword`.

rules/evasion/fake-process-name.yara
egibs: Good point -- I was overthinking this. I'll just ignore `DataDog` and `datadog`.

tstromberg (approved):

rules/evasion/fake-process-name.yara
egibs: Both `DataDog` and `datadog` were present so I used a small expression to cover both cases in [`d34ecb7` (#300)](https://github.com/chainguard-dev/bincapz/pull/300/commits/d34ecb7b8a48db5055013de6c39b39e59db6eab3).

egibs:
```
/bincapz # yara rules/evasion/fake-process-name.yara /usr/bin/process-agent -s
/bincapz # yara rules/evasion/fake-process-name.yara /usr/bin/process-agent -c
0
```

Issue #285, opened 2024-06-24
ref/path/dev/shm critical for byobu-6.12/usr/bin/vigpg: /dev/shm/.vigpg-XXXXXXXXXXXX


tstromberg:
This alert should probably be downgraded to high, at least in the case where there is XXXXX in the filename?

```
/home/t/packages/x86_64/byobu-6.12/usr/bin/vigpg [🚨 CRITICAL]
--------------------------------------------------------------------------------------------------------
RISK  KEY               DESCRIPTION                                       EVIDENCE                      
--------------------------------------------------------------------------------------------------------
CRIT  ref/path/dev/shm  path reference within /dev/shm (world writeable)  /dev/shm/.vigpg-XXXXXXXXXXXX  
--------------------------------------------------------------------------------------------------------
```

PR #302, opened 2024-06-29
Add separate rule for mkstemp paths in /dev/shm

egibs:
Closes: #285 

This PR adds a quick check to see if the match string for the `dev_shm_hidden` contains a hidden path with two or more consecutive `X` characters (we can make this higher if necessary).

Before:
```
/bincapz # go run . ./byobu-6.12/usr/bin/vigpg
byobu-6.12/usr/bin/vigpg [🚨 CRITICAL]
---------------------------------------------------------------------------------------------------------------
RISK  KEY                      DESCRIPTION                                       EVIDENCE
---------------------------------------------------------------------------------------------------------------
NONE  mem/lock                 lock a processes virtual address space            mlock
NONE  ref/path/dev/null        References /dev/null                              /dev/null
LOW   env/TERM                 Look up or override terminal settings             TERM
LOW   fs/file/delete/forcibly  Forcibly deletes files using rm                   rm -f
LOW   fs/tempfile/create       Uses mktemp to create temporary files             mktemp
LOW   ref/path/hidden          hidden path in a system directory                 /dev/shm/.vigpg-XXXXXXXXXXX
LOW   ref/site/url             contains embedded HTTP URLs                       http://www.gnu.org/licenses/
MED   admin/package/install    install software with apt-get                     apt-get install gnupg
MED   ref/program/sudo         calls sudo                                        sudo apt-get install gnupg
MED   shell/exec               executes shell                                    /bin/sh
MED   shell/ignore_output      Runs shell commands but throws output away        which gpg >/dev/null 2>&1
CRIT  ref/path/dev/shm         path reference within /dev/shm (world writeable)  /dev/shm/.vigpg-XXXXXXXXXXXX
---------------------------------------------------------------------------------------------------------------
```

After:
```
/bincapz # go run . ./byobu-6.12/usr/bin/vigpg
byobu-6.12/usr/bin/vigpg [🔥 HIGH]
---------------------------------------------------------------------------------------------------------------
RISK  KEY                      DESCRIPTION                                       EVIDENCE
---------------------------------------------------------------------------------------------------------------
NONE  mem/lock                 lock a processes virtual address space            mlock
NONE  ref/path/dev/null        References /dev/null                              /dev/null
LOW   env/TERM                 Look up or override terminal settings             TERM
LOW   fs/file/delete/forcibly  Forcibly deletes files using rm                   rm -f
LOW   fs/tempfile/create       Uses mktemp to create temporary files             mktemp
LOW   ref/path/hidden          hidden path in a system directory                 /dev/shm/.vigpg-XXXXXXXXXXX
LOW   ref/site/url             contains embedded HTTP URLs                       http://www.gnu.org/licenses/
MED   admin/package/install    install software with apt-get                     apt-get install gnupg
MED   ref/program/sudo         calls sudo                                        sudo apt-get install gnupg
MED   shell/exec               executes shell                                    /bin/sh
MED   shell/ignore_output      Runs shell commands but throws output away        which gpg >/dev/null 2>&1
HIGH  ref/path/dev/shm         reference file within /dev/shm (world writeable)  /dev/shm/.vigpg-XXXXXXXXXXXX
---------------------------------------------------------------------------------------------------------------
```


rules/ref/path/dev-shm.yara
```diff
@@ -15,7 +15,7 @@ rule dev_shm_file : high {
     hash_2023_BPFDoor_8b9d = "8b9db0bc9152628bdacc32dab01590211bee9f27d58e0f66f6a1e26aea7552a6"
     hash_2023_OK_ad69 = "ad69e198905a8d4a4e5c31ca8a3298a0a5d761740a5392d2abb5d6d2e966822f"
   strings:
-    $ref = /\/dev\/shm\/[\%\w\.]{2,64}/
+    $ref = /\/dev\/shm\/[\%\w\.\-\/]{2,64}/
```
egibs, line 18: This was cutting off the full result (e.g., `/dev/shm/.vigpg-XXXXXXXXXXXX` was being truncated to `/dev/shm/.vigpg`).

pkg/report/report.go
```diff
@@ -182,6 +182,18 @@ func behaviorRisk(ns string, rule string, tags []string) int {
 	for _, tag := range tags {
 		if r, ok := levels[tag]; ok {
 			risk = r
+			break
+		}
+	}
+
+	// https://github.com/dustinkirkland/byobu/blob/cd253f0229b68d3c2cffa8d862309c6ea7019d94/usr/bin/vigpg#L25
```
tstromberg, line 189: Let's keep all rule-specific behaviors in YARA. Here's how I would probably set out to solve it:

- Exclude this expression (effectively mkstemp(3)-style use) from the current YARA rule
- Add a new rule specifically for mkstemp(3) usage of /dev/shm with a lower priority.


Probably the best way to do this is up this regexp as an exclusion for the existing rule, and then add another rule that specifically matches

egibs: Cool -- that makes sense!

egibs: Updated in [`f1b2733` (#302)](https://github.com/chainguard-dev/bincapz/pull/302/commits/f1b27330e8325dff10f09ff410902b58d4cf15d2).

rules/ref/path/dev-shm.yara
```diff
@@ -39,7 +39,17 @@ rule dev_shm_hidden : critical {
...
     hash_2023_Unix_Downloader_Rocke_228e = "228ec858509a928b21e88d582cb5cfaabc03f72d30f2179ef6fb232b6abdce97"
   strings:
-    $dev_shm = /\/dev\/shm\/\.[%\w\.\-\/]{0,64}/
+    $dev_shm = /\/dev\/shm\/\.[\%\w\.\-\/]{0,64}/
+    $ignore_mkstemp = /\/dev\/shm\/\.[%\w\.\-\/]{0,64}X{6}/
+  condition:
+    $dev_shm and not $ignore_mkstemp
+}
+
+rule dev_shm_hidden_mkstemp : high {
```
tstromberg, line 48: Let's call this a medium for now, at least until we see some malware abuse.

Also, let's change the rule to match all mkstemp style usage, not just hidden ones. We can split them up later if we need to elevate risk for one or the other.

rules/ref/path/dev-shm.yara
```diff
@@ -39,7 +39,17 @@ rule dev_shm_hidden : critical {
...
+  meta:
+    description = "mkstemp path reference within /dev/shm (world writeable)"
```
tstromberg, line 50: it's worth noting that this is a hidden reference vs a normal one

rules/ref/path/dev-shm.yara
```diff
@@ -39,7 +39,17 @@ rule dev_shm_hidden : critical {
     hash_2023_OrBit_f161 = "f1612924814ac73339f777b48b0de28b716d606e142d4d3f4308ec648e3f56c8"
     hash_2023_Unix_Downloader_Rocke_228e = "228ec858509a928b21e88d582cb5cfaabc03f72d30f2179ef6fb232b6abdce97"
   strings:
-    $dev_shm = /\/dev\/shm\/\.[%\w\.\-\/]{0,64}/
+    $dev_shm = /\/dev\/shm\/\.[\%\w\.\-\/]{0,64}/
+    $ignore_mkstemp = /\/dev\/shm\/\.[%\w\.\-\/]{0,64}X{6}/
```
tstromberg, line 43: can this be updated to ignore all mkstemp refs, not just hidden ones?

tstromberg (changes requested):

tstromberg (approved):

Issue #267, opened 2024-06-10
awesome-yara Rules


egibs:
We've drawn already drawn from several of the rulesets, but I wanted to post the full `awesome-yara` list for easy reference: https://github.com/InQuest/awesome-yara/blob/master/README.md#rules

There are 68 different sources listed of which we've sourced 3-4; though, we've also sourced rules not included on this list -- obviously, not all 68 are relevant but there's an abundance of content here.

tstromberg:
Closing this as we haven't found anything to indicate that it would supply enough accurate coverage to be worth the extra CPU time.

PR #304, opened 2024-06-29
Add Florian Roth's rules

egibs:
Relates to: #267 

Florian Roth's [signature-base](https://github.com/Neo23x0/signature-base) repository contains a lot of interesting rules.

TBD how much overlap there is with what we currently have but I wanted to get a PR going to add them -- we'll also want to ensure that YaraCI doesn't have any false negatives.

The repository is licensed under DRL 1.1 which allows us to integrate these rules so long as we attribute them appropriately (which will happen automatically via Rule metadata).


egibs:
The `Lint` Check isn't having the best time:
```
reviewdog: post failed for misspell: fail to parse diff: GET https://api.github.com/repos/chainguard-dev/bincapz/pulls/304: 406 Sorry, the diff exceeded the maximum number of files (300). Consider using 'List pull requests files' API or locally cloning the repository instead. [{Resource:PullRequest Field:diff Code:too_large Message:}]
```

pkg/action/scan.go
```diff
@@ -58,6 +58,8 @@ func findFilesRecursively(ctx context.Context, root string, c Config) ([]string,
 
 // cleanPath removes the temporary directory prefix from the path.
 func cleanPath(path string, prefix string) string {
+	// if the path has "private" prefix, remove it
+	path = strings.ReplaceAll(path, "/private", "")
```
egibs, line 62: Not sure why this started happening (on MacOS) but the `path` had a `/private/...` prefix which caused the `TrimPrefix` to return the original path unaltered.

third_party/yara/signature_base/configured_vulns_ext_vars.yar
```diff
@@ -1,241 +0,0 @@
...
-		$command12 = "/vim " ascii
-		$command13 = "/rvi " ascii
-		$command14 = "/rvim " ascii
-		//$command15 = "/more " ascii
-		$command16 = "/less " ascii
-		$command17 = "/dd " ascii
-		/* $command18 = "/mount " ascii prone to FPs */
-
-	condition:
-		( filename == "sudoers" or filepath contains "/etc/sudoers.d" ) and
```
egibs: The `filename` and `filepath` external variables were invalid for about 12 of the copied rule files so I excluded those from the copy when running `update.sh`.

tstromberg:
Since Florian is also the guy behind YARAForge, I suspect the best of his rules are already part of bincapz. Is there anything you feel like is missing?

egibs:
> Since Florian is also the guy behind YARAForge, I suspect the best of his rules are already part of bincapz. Is there anything you feel like is missing?

That's a good question. I'll comb through them and see if there's something more targeted we can implement. I'll carry over the path formatting fix into a new PR.

egibs:
Closing for now. Cherry-picking specific rules and adding samples for them is more useful than adding all of the rules (some of which may cause false positives or need to be turned off).

Issue #289, opened 2024-06-24
probable false: net/ddos in datadog-agent-oci-compat-7.54 (synflood)


tstromberg:
```
packages/x86_64/datadog-agent-oci-compat-7.54/opt/datadog-agent/embedded/share/system-probe/ebpf/co-re/tracer-fentry-debug.o [🚨 CRITICAL]
---------------------------------------------------------------------------------
RISK  KEY       DESCRIPTION                                            EVIDENCE  
---------------------------------------------------------------------------------
CRIT  net/ddos  Performs DDoS (distributed denial of service) attacks  synflood  
---------------------------------------------------------------------------------
```

egibs:
After a bit of `xxd` wrangling:
```
000334a0: 0110010001100101 0110011001100101 0111001001011111 0110000101100011 0110001101100101 0111000001110100 0000000001110011 0111100101101110 0110011001101100 0110111101101111 0110010001011111 0111011101100001 0111001001101110 0110010101100100 0000000001111001 0110111101110101  defer_accept.synflood_warned.you
```

We can add a `not` for this 👍🏻

tstromberg:
I'm still seeing this with wolfi datadog packages & bincapz 0.13.2:

```
/home/t/packages/x86_64/datadog-agent-oci-compat-7.54/opt/datadog-agent/embedded/share/system-probe/ebpf/co-re/tcp-queue-length-debug.o [
🚨 CRITICAL]
---------------------------------------------------------------------------------
RISK  KEY       DESCRIPTION                                            EVIDENCE  
---------------------------------------------------------------------------------
CRIT  net/ddos  Performs DDoS (distributed denial of service) attacks  synflood  
---------------------------------------------------------------------------------

/home/t/packages/x86_64/datadog-agent-oci-compat-7.54/opt/datadog-agent/embedded/share/system-probe/ebpf/co-re/tcp-queue-length.o [🚨 CRIT
ICAL]
---------------------------------------------------------------------------------
RISK  KEY       DESCRIPTION                                            EVIDENCE  
---------------------------------------------------------------------------------
CRIT  net/ddos  Performs DDoS (distributed denial of service) attacks  synflood  
---------------------------------------------------------------------------------

/home/t/packages/x86_64/datadog-agent-oci-compat-7.54/opt/datadog-agent/embedded/share/system-probe/ebpf/co-re/tracer-debug.o [🚨 CRITICAL
]
---------------------------------------------------------------------------------
[...]

PR #316, opened 2024-07-02
Split up the ddos_refs rule for improved accuracy

tstromberg:
With this change, merely mentioning `synflood` is only a medium, not a critical.

While we solved datadog, I encountered other software packages that mentioned `synflood` in a healthy way, so I've split the queries up so that it's unlikely to have a critical hit any longer.


rules/net/ddos.yara
```diff
@@ -1,15 +1,26 @@
...
-    $ignore_ref = "synflood_warned"
+    $tsource = "TSource Engine Query"
   condition:
-    any of ($ref*) and not $ignore_ref
+    $tsource
+}
+
+rule tcp_flood_refs : high {
+  meta:
+    description = "May performs DDoS (distributed denial of service) attacks"
```
egibs, line 15:
```suggestion
    description = "May perform DDoS (distributed denial of service) attacks"
```

egibs (approved):

egibs:
Closes: #289

PR #323, opened 2024-07-04
Fix IgnoreSelf behavior

### LINKED ISSUE #317: bincapz detects bincapz binaries as malware (--ignore-self doesn't work)
I noticed this while looking at false-positives for wolfi:

```
$ go install .
$ go run . --min-risk=critical $HOME/go/bin/bincapz
```

generates pages of output:

```                                                                                                             tunnel_out_ssh                                      
CRIT  tools/net/venom                    Uses Venom, a multi-hop proxy and RAT for penetration               /Venom/agent                                        
                                         testers                                                             Dliv3/Venom                                         
                                                                                                             venom_agent                                         
CRIT  tools/recon/pspy                   tool to snoop on processes without root permissions                 dominicbreuker/pspy                                 
                                                                                                             startFSW                                            
                                                                                                             startPSS                                            
                                                                                                             triggerEvery                                        
CRIT  tools/vulncheck/metasploit         Metasploit shellcode (msfpayload)                                   /shh/bin                                            
                                                                                                             Payload:                                            
                                                                                                             http://www.metasploit.com                           
                                                                                                             msfpayload                                          
```

egibs:
Closes #317

`IgnoreSelf` was relying on `path == self` in order to work correctly.

Here's an example of `path` and `self` inside of Wolfi, respectively:
```
/bincapz # go run . --ignore-self=true /usr/bin/bincapz
/usr/bin/bincapz /tmp/go-build1161677753/b001/exe/bincapz
^ path           ^ self
```

This PR tweaks the `IgnoreSelf` logic and consolidates it within the `Generate` function in `pkg/report.go` and also adds a test to validate that `bincapz` is ignored as expected.


egibs:
We could keep the `path == self` comparison in addition to this new logic but it would be redundant for cases where `/usr/bin/bincapz` scans `/usr/bin/bincapz`.

pkg/action/scan.go
```diff
@@ -45,7 +44,7 @@ func findFilesRecursively(ctx context.Context, root string, c Config) ([]string,
 				return nil
 			}
 
-			if c.IgnoreSelf && path == self {
+			if c.IgnoreSelf && strings.Contains(path, BINARY) {
```
tstromberg, line 47: I think we have multiple forms of broken behavior going on; each with their own fragility: path based and YARA based (https://github.com/chainguard-dev/bincapz/blob/main/rules/internal/bincapz.yara).

What do you think about us taking a strict approach that only ignores the file if:

- The file basename is `bincapz`
- The YARA rule matches

tstromberg: The concern I have is someone reporting a "vulnerability" that they can just name their malware `bincapz` and get around the checks. At least this way they'll have to work for it.

egibs: Very true -- I'll work on those improvements.

egibs: Updated in [`deb9987` (#323)](https://github.com/chainguard-dev/bincapz/pull/323/commits/deb9987abba39488fbe5c48910b62d7cd8d259bf).

I tried to consolidate all of the logic in `Generate` -- the downside here is that the binary will be scanned regardless of whether we're ignoring it so it won't be as fast as skipping over it when adding it to the list of files to scan.

pkg/report/report.go
```diff
@@ -303,18 +307,24 @@ func Generate(ctx context.Context, path string, mrs yara.MatchRules, ignoreTags
...
+		}
 		risk = behaviorRisk(m.Namespace, m.Rule, m.Tags)
 		if risk > overallRiskScore {
 			overallRiskScore = risk
 		}
 		riskCounts[risk]++
-		if risk < minScore {
+		// The bincapz rule is classified as harmless
+		// This will prevent the rule from being filtered
+		if risk < minScore && !ignoreBincapz {
```
egibs, line 327: I spent way too long trying to figure out why we never got to the `__bincapz__` case statement. Since the `bincapz` rule is harmless and the default risk is `low`, the rule was being filtered.

pkg/report/report.go
```diff
@@ -353,8 +363,10 @@ func Generate(ctx context.Context, path string, mrs yara.MatchRules, ignoreTags
 				}
 			case "author_url":
 				b.RuleAuthorURL = v
-			case "__bincapz__":
-				fr.IsBincapz = true
+			case fmt.Sprintf("__%s__", BINARY):
+				if v == "true" {
```
egibs, line 367: Just a quick check to ensure that the metadata value is actually `"true"` instead of solely relying on the presence of the `__bincapz__` identifier.

rules/internal/bincapz.yara
```diff
@@ -3,9 +3,14 @@ rule bincapz : harmless {
...
+		$f_action = "bincapz/pkg/action"
+		$f_bincapz = "bincapz/pkg/bincapz"
+		$f_compile = "bincapz/pkg/compile"
+		$f_profile = "bincapz/pkg/profile"
+		$f_render = "bincapz/pkg/render"
+        $f_report = "bincapz/pkg/report"
+        $f_version = "bincapz/pkg/version"
     condition:
-        2 of them
+        $b_behavior and 2 of ($f_*)
```
egibs, line 15: Might make sense to do `all of ($f_*)`?

egibs: Updated in [`80f7ed1` (#323)](https://github.com/chainguard-dev/bincapz/pull/323/commits/80f7ed166eefce38f72557f101b24222c396f6a0).

rules/internal/bincapz.yara
```diff
@@ -3,9 +3,14 @@ rule bincapz : harmless {
...
+		$f_action = "bincapz/pkg/action"
+		$f_bincapz = "bincapz/pkg/bincapz"
+		$f_compile = "bincapz/pkg/compile"
+		$f_profile = "bincapz/pkg/profile"
+		$f_render = "bincapz/pkg/render"
+        $f_report = "bincapz/pkg/report"
+        $f_version = "bincapz/pkg/version"
     condition:
-        2 of them
+        $b_behavior and all of ($f_*)
```
tstromberg, line 15: Requiring every f* will make this fragile when testing against older and newer bincapz releases. I'd change it back or switch it to 3 of them.

egibs: Updated in [`f670fbc` (#323)](https://github.com/chainguard-dev/bincapz/pull/323/commits/f670fbcc1c263d38e06f5b02cb9f44bf15e9d0b7).

pkg/action/scan_test.go
```diff
@@ -128,3 +138,42 @@ func TestFormatPath(t *testing.T) {
 		})
 	}
 }
+
+func TestBincapzIgnored(t *testing.T) {
```
egibs, line 142:
```
$ go test -run "^TestBincapzIgnored\$" github.com/chainguard-dev/bincapz/pkg/action
ok  	github.com/chainguard-dev/bincapz/pkg/action	35.098s
```

pkg/render/simple.go
```diff
@@ -24,8 +24,10 @@ func (r Simple) File(_ context.Context, fr *bincapz.FileReport) error {
 	if fr.Skipped != "" {
 		return nil
 	}
-
-	fmt.Fprintf(r.w, "# %s\n", fr.Path)
```
egibs, line 28: This was displaying an empty `#`:
```
# 
```

tstromberg (approved):
Sweet! Thank you for fixing this!

PR #326, opened 2024-07-04
Detect Mallox, Ladvik & NPM droppers

tstromberg:
Added rules for a smattering of malware I've been reading up on recently.

I needed to add `.json` as a valid program extension, as that's what NPM uses for preinstall/postinstall.

NOTE: Some rules were reformatted by a linter (`yato`) when I used it to add hashes. Some samples were dropped because EDR software doesn't like ransomware.


tstromberg:
Damned EDR software deleting files before they can get committed :(

egibs (approved):

PR #328, opened 2024-07-05
Drastically improve detection for NPM registry malware

tstromberg:
Results of an initial audit to see how well we detect malware from the Backstabbers-Knife-Collection NPM repository. This adds enough rules to detect the first 20+ samples I reviewed, but there is still a long way to go.

While most of the improvements are focused on NPM and NodeJS, some rules are universally improved, such as the browser credential stealer rules.

NOTE: This also fixes a minor bug in `refresh-testdata.sh` which would corrupt any samples that had a `.json` extension.


rules/kernel/platform.yara
```diff
@@ -67,9 +67,14 @@ rule npm_uname : medium {
...
+    $ = "process.platform"
+    $ = "process.arch"
+    $ = "process.versions"
+    $ = "os.platform()"
+    $ = "os.arch()"
+    $ = "os.release()"
   condition:
     any of them
 }
+
```
egibs, line 79: Nit: two empty lines here

rules/user/lookup.yara
```diff
@@ -1,14 +0,0 @@
-rule getpwuid : harmless {
```
egibs, line 1: Were these rules supposed to be added to `rules/user/current.yara`?

rules/random/bytes.yara
```diff
@@ -0,0 +1,10 @@
+
+rule rand {
+	meta:
+		description = "generates random bytes"
+	strings:
+		$ref = ".randomBytes(" fullword
+	condition:
+		$ref
+}
+
```
egibs, line 10: Nit: empty line

rules/combo/stealer/browser.yara
```diff
@@ -21,26 +21,39 @@ rule multiple_browser_credentials : high {
...
+	$fs_user_data = "User Data"
+
+	$name_brave = "Brave-Browser"
+    $name_brave_software = "BraveSoftware"
+    $name_chrome = "Google Chrome"
+    $name_opera_gx = "OperaGX"
+	$name_firefox = "Firefox"
+	$name_opera_gx_dir = "Opera GX"
+    $name_opera = "Opera Software"
+	$name_yandex = "YandexBrowser"
```
egibs, line 51: Would it make sense to add Safari or Vivaldi here (DDG and Orion are also possibilities)?

rules/combo/recon/nodejs.yara
```diff
@@ -1,14 +1,53 @@
...
     $proc3 = "process.versions"
-    $h = "http.request"
+	$request_http = /require\([\"\']https{0,1}[\"\']\)/
     $post = "POST"
   condition:
-    filesize < 33554432 and $h and $post and any of ($proc*)
+    filesize < 16KB and $request_http and $post and any of ($proc*)
+}
+
+rule nodejs_phone_home : high {
```
egibs, line 16: Does this need to be private or does that not matter all that much?

rules/combo/stealer/browser.yara
tstromberg: Safari is special as it cookies in `~/Library/Cookies` (covered by $f_cookies)

I haven't seen malware samples that target the others, but if I do, we'll add them here.

rules/combo/recon/nodejs.yara
tstromberg: nope, I still want this rule to show if it matches, even if nothing that relies on it does.

rules/user/lookup.yara
tstromberg: good catch that these were accidentally removed - re-added them.

egibs (approved):

PR #339, opened 2024-07-09
Reduce HIGH rule hits in renovate, nodejs, falco

tstromberg:
Also removes duplicated `site/unusual.yara` rule file.


egibs (approved):
Changes LGTM -- should we add clean samples to validate against at some point?

PR #342, opened 2024-07-09
Improve frequency-based upgrade heuristics, add flag to disable

tstromberg:
This PR improves frequency risk upgrade heuristics by:

* including file size in the calculation
* adding a `--quantity-increases-risk` flag to disable the feature (other flag naming ideas welcome)

Currently, any program with >4 HIGH results is classified as CRITICAL. 

This can cause a lot of problems with really large programs, particularly Go binaries with a lot of dependencies. This PR adds size-based heuristics, but also allows you to turn the feature off entirely, which is ideal for environments such as CI/CD where you care more about capabilities diffs and raw capability risks rather than a synthetic combined values.


pkg/report/report.go
```diff
@@ -465,6 +478,47 @@ func Generate(ctx context.Context, path string, mrs yara.MatchRules, ignoreTags
...
+	// bloated go binaries
+	if sizeMB < 20 && highCount > 4 {
+		upgrade = true
+	}
+
+	if highCount > 6 {
+		upgrade = true
+	}
+
+	if !upgrade {
```
egibs, line 513: Can we just return `upgrade` at the end? If none of the conditions are `true` it will still be `false`.

pkg/report/report_test.go
```diff
@@ -63,38 +64,26 @@ func BenchmarkLongestUnique(b *testing.B) {
...
-			name:       "One True; Many False",
-			conditions: []bool{true, false, false, false},
-			want:       false,
-		},
-		{
-			name:       "Many True; One False",
-			conditions: []bool{true, true, true, false},
-			want:       false,
-		},
+		{"no risk", 0, map[int]int{}, 1024, false},
```
egibs, line 75: Nice 🚀

pkg/report/report.go
tstromberg: This logic was added to prevent extraneous logging for the normal case where no upgrade is done.

egibs (approved):

bincapz.go
```diff
@@ -70,6 +70,7 @@ func main() {
 	errFirstHitFlag := flag.Bool("err-first-hit", false, "exit with error if scan source has matching capabilities")
 	ociFlag := flag.Bool("oci", false, "Scan an OCI image")
 	omitEmptyFlag := flag.Bool("omit-empty", false, "Omit files that contain no matches")
+	frequencyUpgradeFlag := flag.Bool("frequency-upgrade", true, "increase file risk score based on frequency heuristics")
```
egibs, line 73: Maybe `upgrade-severity` is a better flag name?

tstromberg: How about --quantity-increases-risk ?

egibs: I think that works. Since this defaults to `true` it's not something that needs to be written out all that often.

PR #347, opened 2024-07-10
Address more false HIGH results in open-source projects

tstromberg:
The worst of which was a new false - `impossibly_small_elf_program` - that was recently upgraded to HIGH. I had to use the "elf" YARA module to fix it.


egibs:
This looks like an artifact that needs to be addressed:
```
  string(
            - 	"",
            + 	"# Linux/clean/libBrokenLocale-2.27.so\n",
              )
```

egibs (approved):

Issue #348, opened 2024-07-10
Remove unused `--omit-empty` flag


tstromberg:
I believe this flag is broken, and nobody uses it.

I think the intended behavior (files with no results are omitted from the output) is already happening elsewhere.

Related to #347

egibs:
I agree -- we should just passively omit empty files instead of having a separate flag to control it.

PR #350, opened 2024-07-10
Fix empty testdata file diff; fix simple diff bug

egibs:
Closes: https://github.com/chainguard-dev/bincapz/issues/349

This was tricky to track down. The file wasn't being skipped but the `fr.Behaviors` check was invalid because the slice was of length `0` rather than being `nil`:
```
fmt.Println(bool(fr.Behaviors == nil))
fmt.Println(bool(len(fr.Behaviors) == 0))
false
true
```

This PR fixes the bug and also addresses one small `Full` diff bug with the `simple` renderer; that is, `rep.Diff.Removed` was being referenced for added files rather than `rep.Diff.Added`.


samples/macOS/clean/ls.sdiff.level_2
```diff
@@ -1,2 +1,2 @@
-*** changed: macOS/clean/ls
+--- missing: .
```
egibs, line 1: Hmmm, I'll look into this as well.

pkg/render/markdown.go
```diff
@@ -42,7 +42,9 @@ func matchFragmentLink(s string) string {
 }
 
 func (r Markdown) File(ctx context.Context, fr *bincapz.FileReport) error {
-	markdownTable(ctx, fr, r.w, tableConfig{Title: fmt.Sprintf("## %s [%s]", fr.Path, mdRisk(fr.RiskScore, fr.RiskLevel))})
+	if len(fr.Behaviors) != 0 {
```
tstromberg, line 45: NOTE: This further breaks `--omit-empty`, but we should get rid of it anyways, IMHO (see #348)

samples/macOS/clean/ls.sdiff.level_2
egibs: Resolved in [`0b94a09` (#350)](https://github.com/chainguard-dev/bincapz/pull/350/commits/0b94a097a4413b90e77b234b3560fe3c4d2917be).

samples/macOS/clean/ls.sdiff.trigger_2
```diff
@@ -1,5 +1,6 @@
-*** changed: macOS/clean/ls
```
tstromberg, line 1: I don't understand this output change either.

samples/macOS/clean/ls.sdiff.level_2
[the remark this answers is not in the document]
egibs: That was an unintended side-effect of my risk level check. I removed it and will instead rely on the `len(fr.Behaviors)...` check.

tstromberg (approved):

PR #351, opened 2024-07-10
Remove OmitEmpty; fix nil pointer dereference

egibs:
Closes: https://github.com/chainguard-dev/bincapz/issues/348

This PR removes the [unused] `--omit-empty` flag. 

Additionally, since we're now returning `nil` when only `len(fr.Behaviors)` is `0`, the ordering of `if fr.Error != "" { ... }` before `if fr == nil { ... }` was causing a nil pointer dereference so I fixed that as well.

### DIFF
--- a/pkg/action/scan.go
+++ b/pkg/action/scan.go
@@ -102,7 +102,7 @@ func scanSinglePath(ctx context.Context, c bincapz.Config, yrs *yara.Rules, path
 		fr.Path = fmt.Sprintf("%s ∴ %s", absPath, formatPath(cleanPath))
 	}
 
-	if len(fr.Behaviors) == 0 && c.OmitEmpty {
+	if len(fr.Behaviors) == 0 {
 		return nil, nil
 	}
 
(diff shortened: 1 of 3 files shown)


egibs:
Looking into this again:
```
--- FAIL: TestDiff/macOS/clean/ls.sdiff.level_2 (0.05s)
        samples_test.go:232: simple diff output mismatch: (-want +got):
              (
              	"""
            - 	*** changed: macOS/clean/ls
            + 	--- missing: .
              	-process/name/set
              	"""
              )
```

tstromberg (approved):

pkg/action/scan.go
```diff
@@ -102,8 +102,8 @@ func scanSinglePath(ctx context.Context, c bincapz.Config, yrs *yara.Rules, path
 		fr.Path = fmt.Sprintf("%s ∴ %s", absPath, formatPath(cleanPath))
 	}
 
-	if len(fr.Behaviors) == 0 && c.OmitEmpty {
-		return nil, nil
+	if len(fr.Behaviors) == 0 {
+		return &bincapz.FileReport{Path: path}, nil
```
egibs, line 106: Without this, `rel, err := filepath.Rel(fromPath, f.Path)` results in `.`.

PR #355, opened 2024-07-11
Ignore SIGNATURE_BASE_SUSP_PS1_JAB_Pattern_Jun22_1 rule

egibs:
This rule is causing false positives: https://github.com/wolfi-dev/os/pull/23799

This PR turns the rule off (since it's a third-party rule we can't modify it directly); however, we can potentially roll our own rule for this in the future and add exclusions.


pkg/compile/compile.go
```diff
@@ -20,55 +20,56 @@ var FS = rules.FS
 // badRules are noisy 3rd party rules to silently disable.
 var badRules = map[string]bool{
 	// YARAForge
+	"GCTI_Sliver_Implant_32Bit":          true,
 	"GODMODERULES_IDDQD_God_Mode_Rule":   true,
 	"MALPEDIA_Win_Unidentified_107_Auto": true,
-	"GCTI_Sliver_Implant_32Bit":          true,
+	"SIGNATURE_BASE_SUSP_PS1_JAB_Pattern_Jun22_1", true,
```
egibs, line 26: This is the only addition.

The rest of the changes are from sorting all of the lines.

tstromberg (approved):

PR #356, opened 2024-07-12
Allow move calculations for all files

egibs:
Given how slow move calculations can be, only `.so` library files have been considered when running bincapz with the `--diff` flag. 

This PR allows for all files to be considered during move calculations and addresses the `missing`/`added` output seen here: https://github.com/wolfi-dev/os/pull/23820

I added a quick diff test for this case and the output is now:
```
*** changed: Linux/clean/aws-c-io/aws-c-io-0.14.11-r0.spdx.json
```

This PR adds a function that flattens the diff report maps into a single slice so that iterating through them and calculating scores is faster.


hectorj2f (approved):

samples/Linux/2023.FreeDownloadManager/freedownloadmanager.sdiff
```diff
@@ -1,7 +1,17 @@
-*** changed: Linux/2023.FreeDownloadManager/freedownloadmanager_infected_postinst
+--- missing: Linux/2023.FreeDownloadManager/freedownloadmanager_clear_postinst
+-admin/add_apt_key
+-data/embedded/pgp/key
+-net/download
+-ref/path/etc
+-ref/path/usr/bin
+-ref/site/url
+-shell/ignore_output
+++++ added: Linux/2023.FreeDownloadManager/freedownloadmanager_infected_postinst
```
tstromberg, line 9: This should be a file change, not a add/removal.

samples/macOS/clean/ls.mdiff
```diff
@@ -1,16 +1,18 @@
-## Changed: . [⚠️ MEDIUM → ✅ LOW]
```
tstromberg, line 1: This should be a file change - but instead of `.`, it should say the name of the new filename.

egibs: Addressed in [`dfc5173` (#356)](https://github.com/chainguard-dev/bincapz/pull/356/commits/dfc517390378280a98a62403b6d9c29f3d6075e3).

tstromberg (approved):

egibs:
Still working on the `moved` logic for the new sample but once that's done I'll merge this.

samples/Linux/2023.FreeDownloadManager/freedownloadmanager.sdiff
```diff
@@ -1,4 +1,4 @@
-*** changed: Linux/2023.FreeDownloadManager/freedownloadmanager_infected_postinst
+>>> moved: Linux/2023.FreeDownloadManager/freedownloadmanager_clear_postinst -> Linux/2023.FreeDownloadManager/freedownloadmanager_infected_postinst (score: 0.938235)
```
tstromberg, line 1: When diffing two binaries, it shouldn't be considered a move; just a diff between two programs. 'Moved' should only be considered if directory trees are diffed.

pkg/action/diff.go
```diff
@@ -230,8 +244,10 @@ func fileMove(ctx context.Context, c bincapz.Config, fr, tr *bincapz.FileReport,
 		}
 	}
 
-	// Move these into the modified list.
-	d.Modified[apath] = abs
-	delete(d.Removed, rpath)
-	delete(d.Added, apath)
+	// Move these into the modified list if the files are not completely different (something like ~0.3)
+	if score > 0.3 {
```
egibs, line 243: I added this because running something like this makes bincapz think the original binary was changed:
```sh
$ go run . --format=simple --diff samples/Linux/clean/ls.x86_64 samples/macOS/2024.LightSpy/dropper
```

tstromberg (changes requested):

samples/Linux/2023.FreeDownloadManager/freedownloadmanager.sdiff
egibs: Makes sense.

I'll land the original changes as-is and then spend some more time overhauling the diff behavior. I don't want this scope creep to hold up the next release.

egibs: Moved back to `filepath.Rel` in [`55a9ed2` (#356)](https://github.com/chainguard-dev/bincapz/pull/356/commits/55a9ed242264ffefb9ab919d59fac16d4bde14ae).

tstromberg (approved):

PR #358, opened 2024-07-12
Avoid extracting invalid archive files

egibs:
We've had several instances of `.gz` and `.tar` files failing to extract:
```
$ go run . ~/Downloads/joblib_0.9.4.dev0_compressed_cache_size_pickle_py35_np19.gz
time=2024-07-12T13:52:53.206-05:00 level=ERROR source=/.../.../repos/bincapz/pkg/action/scan.go:199 msg="unable to process /Users/evangibler/Downloads/joblib_0.9.4.dev0_compressed_cache_size_pickle_py35_np19.gz: extract to temp: extract: failed to create gzip reader: gzip: invalid header"
```

This is because the files are actually data files and not archives:
```
/.../.../Downloads/joblib_0.9.4.dev0_compressed_cache_size_pickle_py35_np19.gz: data
```

Additionally, [this](https://github.com/mafintosh/tar-fs/blob/master/test/fixtures/invalid.tar) file was being extracted and is also invalid. Luckily, its type (`palmpilot database/document file`) is not covered by our case statement so we can rely on the same `programKind(ctx, path) != ""` check to avoid that as well.


egibs:
I have some tests to fix 👀

egibs:
> I have some tests to fix 👀

Fixed in [`ceb7ed8` (#358)](https://github.com/chainguard-dev/bincapz/pull/358/commits/ceb7ed894e02ab61cab500d446bd66f880b77f65)

pkg/action/programkind.go
```diff
@@ -78,22 +79,37 @@ func programKind(ctx context.Context, path string) string {
...
 
-	// TODO: Is it safe to log unsanitized file stuff?
-	logger.Debug("magic", slog.String("desc", desc), slog.String("header", headerString), slog.Any("err", err))
-
 	if found, kind := byExtension(path); found {
+		// if we get a map hit, check if we're looking at an archive and that it has a common archive header
+		if _, ok := archiveMap[getExt(path)]; ok {
+			h := header[:n]
+			switch {
+			case bytes.HasPrefix(h, []byte("PK\x03\x04")) ||
```
tstromberg, line 97: If we're going to do this kind of sniffing, we should import a better `magic` detection library instead of implementing it ourselves. I avoided doing so initially because they required cgo, but YARA does as well so it's no loss.

tstromberg (changes requested):
I like the idea of clarifying the error message when an invalid .gz file is encountered, but I don't think the error message should be hidden entirely from users by default. If bincapz is unable to process an archive file, I'd rather know that it can't be processed than assume it's safe.

If nothing else, it'd be nice to have something that says "%s does not appear to be a valid .Z file, won't try to extract ..."; but at that point, you have to wonder if we've gained anything from this added complexity versus just trying and reporting an error message if it isn't possible.

pkg/action/programkind.go
egibs: Looks like `h2non/filetype` may be a good option: https://pkg.go.dev/github.com/h2non/filetype#IsArchive

pkg/action/programkind.go
```diff
@@ -78,22 +79,37 @@ func programKind(ctx context.Context, path string) string {
...
+				bytes.HasPrefix(h, []byte("\x1F\x8B\x08")) ||
+				bytes.HasPrefix(h, []byte{0x75, 0x73, 0x74, 0x61, 0x72}) ||
+				bytes.HasPrefix(h, []byte("\xFD7zXZ\x00")) ||
+				bytes.HasPrefix(h, []byte("ustar")):
```
tstromberg, line 101: I'm pretty sure we're going to find this list woefully incomplete; but if you really wanted to do something like this, it should be as a self-documenting data structure like a dictionary. I can't tell what half of these are from memory.

PR #364, opened 2024-07-15
Fix GitLab healthcheck script false positive

egibs:
This PR addresses the critical false positive seen here: https://github.com/wolfi-dev/os/pull/23899

We can ignore the combination of the comment, `GET` and localhost 8080 strings to ignore this.

I added a sample for this as well.


samples/Linux/clean/healthcheck.simple
```diff
@@ -0,0 +1,5 @@
+# Linux/clean/healthcheck
+net/http/request
+ref/path/dev
+shell/bash_dev_tcp
```
egibs, line 4: This is the `bash_dev_tcp` rule rather than `bash_dev_tcp_hardcoded_ip`.

hectorj2f (approved):

PR #368, opened 2024-07-17
Address false positives for Chromium, KOTS, and Nuclei

egibs:
This PR addresses critical false positives seen here:
- https://github.com/wolfi-dev/os/pull/24105
- https://github.com/wolfi-dev/os/pull/24107
- https://github.com/wolfi-dev/os/pull/24139

Notes:
- The Chromium false positives were due to our only ignoring `Chrome` strings
- The KOTS false positives were due to PHP tags located in a comment
- The Nuclei false positives were due to strings matching Wapalyzer technology JSON content
  - Specifically, content from this file: https://github.com/Lissy93/wapalyzer/blob/main/src/technologies/c.json

I used slightly more specific strings when ignoring the Nuclei false positives since they were originating from JSON content and I wanted to match the exact fields rather than doing something like `$not_description = "\"description\": \""`.

### DIFF
--- a/rules/combo/stealer/wallet.yara
+++ b/rules/combo/stealer/wallet.yara
@@ -24,6 +24,13 @@ rule crypto_stealer : critical {
     $w_trezor = "Trezor"
     $w_exodus = "Exodus"
     $w_coinomi = "Coinomi"
+    $not_cats = /\"cats\": \[[^]]{0,64}/
+    $not_description = /\"description\": "([^"]{0,64})"/
+    $not_dom = /\"dom\": "([^"]{0,64})"/
+    $not_icon = /\"icon\": "([^"]{0,64})"/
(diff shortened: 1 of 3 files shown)


rules/combo/stealer/wallet.yara
```diff
@@ -24,6 +24,13 @@ rule crypto_stealer : critical {
     $w_trezor = "Trezor"
     $w_exodus = "Exodus"
     $w_coinomi = "Coinomi"
+    $not_cats = /\"cats\": \[[^]]{0,64}/
```
egibs, line 27: This isn't an exhaustive representation of the Wapalyzer schema but these should be broad enough to match content from those technology files.

hectorj2f (approved):
LGTM, thanks

PR #375, opened 2024-07-22
Address false positives for SQLPad and Lerna

egibs:
This PR addresses false positives seen here:
- https://github.com/wolfi-dev/os/pull/24378
- https://github.com/wolfi-dev/os/pull/24385

The main issue was the `base64_python_functions` rule triggering on TypeScript code so I added several common JS strings that we can ignore which cleared up the match strings.

The other critical was a third-party rule which was matching all three of these strings for a WASM runtime file (`/usr/local/lib/node_modules/lerna/node_modules/@napi-rs/wasm-runtime/dist/runtime.js`): 
```yara
strings:
    $wasm = "WebAssembly.Module" base64
    $int = "WebAssembly.Instance" base64
    $inst = "WebAssembly.instantiate" base64
```


hectorj2f (approved):
I'd be good to have tests for these

Issue #309, opened 2024-07-02
probable false: combo/stealer/password in caddy 


tstromberg:
With bincapz 0.13.2:

```
/home/t/packages/x86_64/caddy-2.7/usr/bin/caddy [🚨 CRITICAL]
---------------------------------------------------------------------------------------------------------------------
RISK  KEY                     DESCRIPTION                                  EVIDENCE                                  
---------------------------------------------------------------------------------------------------------------------
CRIT  combo/stealer/password  Password finder/dumper, such as MimiPenguin  Password                                  
                                                                           apache2.conf                              
                                                                           exphttp.error.idtrusted_proxyhash-passw…  
                                                                           sshd_config                               
---------------------------------------------------------------------------------------------------------------------

/home/t/packages/x86_64/caddy-2.8/usr/bin/caddy [🚨 CRITICAL]
---------------------------------------------------------------------------------------------------------------------
RISK  KEY                     DESCRIPTION                                  EVIDENCE                                  
---------------------------------------------------------------------------------------------------------------------
CRIT  combo/stealer/password  Password finder/dumper, such as MimiPenguin  Password                                  
                                                                           apache2.conf                              
                                                                           exphttp.error.idtrusted_proxyhash-passw…  
                                                                           sshd_config                               
---------------------------------------------------------------------------------------------------------------------
```

Related: #232

PR #378, opened 2024-07-24
Address false positives for remaining public packages

egibs:
This PR addresses false positives seen with the following packages:
- `caddy`
- `datadog-agent`
- `opa`
- `php`
- `rstudio`
- `sonarqube`
- `varnish`

I also added samples where possible so that we have something to test against.


hectorj2f (approved):

egibs:
Just saw a false positive for `opa` show up so I'll roll that into this PR as well.

hectorj2f (approved):

egibs:
We have a full public package scan running right now so I'll wait for that to finish before merging this in case anything else shows up.

egibs:
Closes: #309

PR #387, opened 2024-07-30
Address false positives for mlflow and pytorch

egibs:
This PR addresses false positives with `mlflow` and `pytorch` -- 
- We already had an exclusion for `mlflow` but it was a little too specific and didn't account for newer versions
- Pytorch installs the `functorch` package which uses `magic_trace.py` which runs a `chmod +x` on the magic trace cache location


hectorj2f (approved):

hectorj2f:
I wonder if we could add some tests for these.

hectorj2f (approved):
Thanks

PR #391, opened 2024-07-30
Address Kibana false positives

egibs:
This PR addresses the false positives seen [here](https://github.com/chainguard-dev/enterprise-packages/pull/6251).

Most of the false positives were from a third-party rule and two were from our rules. One of the latter failures was due to JavaScript code failing a PHP rule and the other was failing due to the strings matching a Kibana security rule.

I've added both of the latter two findings as samples for better coverage.


rules/combo/backdoor/php.yara
```diff
@@ -116,8 +116,13 @@ rule php_eval_gzinflate_base64_backdoor : critical {
     $f_gzinflate = "gzinflate("
     $f_base64_decode = "base64_decode"
 
-    $not_php = "PHP_FLOAT_DIG" fullword
+    $not_js = " ?? "
```
egibs, line 119: We also do this for the `base64_python_functions` rule where JS code was being evaluated in the context of Python code.

rules/secrets/keychain.yara
```diff
@@ -20,26 +22,32 @@ rule macos_library_keychains : medium {
...
 }
 
 rule find_generic_password : high {
   meta:
     description = "Looks up a password from the Keychain"
   strings:
     $ref = /find-generic-passsword[ \-\w\']{0,32}/
-    $ctkcard = "/System/Library/Frameworks/CryptoTokenKit.framework/ctkcard"
+    $not_ctkcard = "/System/Library/Frameworks/CryptoTokenKit.framework/ctkcard"
+    $not_elastic_author = { 22 61 75 74 68 6F 72 22 3A 20 5B 0A 20 20 20 20 22 45 6C 61 73 74 69 63 22 0A 20 20 5D }
```
egibs, line 37:
```json
"author": [
    "Elastic"
  ]
```

hectorj2f (approved):

PR #394, opened 2024-08-05
Bump bincapz version to 0.17.0 ahead of release

egibs:
This PR bumps the bincapz version string to `0.17.0` before cutting the new release.

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -6,7 +6,7 @@ import (
 )
 
 const (
-	ID string = "v0.16.2"
+	ID string = "v0.17.0"
 )
 
 // Check if the build info contains a version.


pkg/version/version.go
```diff
@@ -6,7 +6,7 @@ import (
 )
 
 const (
-	ID string = "v0.16.2"
+	ID string = "v0.17.0"
```
hectorj2f, line 9: Could we set this version based on the tags when building the release? So we avoid hardcoding it.

egibs: We should do that. We're not compiling any binaries as part of the release yet which is a gap in the release process.

hectorj2f: I don't want to block this release but we should change this soon.

hectorj2f (approved):

PR #396, opened 2024-08-05
Store samples in an archived format; add Makefile targets to archive/extract

egibs:
We've hit the ~500MB Go module limit due to the size of our `samples` directory. Long-term, we'll want to handle these as separate modules and work to keep each below 500MB.

For now, we can compress the samples directory (including the test code and sample update script) and then extract the archives locally for testing/updates. Since we're ignoring `samples/` now, we don't have to worry about committing its contents; rather, the main concern is running `make archive-samples` and committing the changes when updating or adding samples.

Additionally, I created multiple 50MB archive files to avoid running afoul of Git's LFS requirement for files > 100MB.

The good news is that we've gone from about ~420MB of content in the `samples/` directory to ~150MB with the archive files:
```
 50M	./samples.tar.gz.aa
 51M	./samples.tar.gz.ab
 47M	./samples.tar.gz.ac
```


Makefile
```diff
@@ -113,3 +113,15 @@ update-third-party:
 .PHONY: refresh-sample-testdata out/bincapz
 refresh-sample-testdata: out/bincapz
 	./samples/refresh-testdata.sh ./out/bincapz
+
+.PHONY: archive-samples
```
hectorj2f, line 117: I'd suggest to add this targets to the `make test` target. So nobody gets an error if they don't run them first.

egibs: Definitely a good idea!

hectorj2f (approved):

PR #402, opened 2024-08-14
Update to go1.22; speed up recursiveScan

egibs:
With Go 1.23.0 being released, I figured it was time for us to update bincapz to 1.22. This PR handles the Go update along with package updates, including `clog` which requires Go 1.22.

I also updated the `slogtest.TestContextWithLogger` calls to `slogtest.Context` in `samples_test.go`.

Additionally, I add some concurrency to `recursiveScan` which makes quite a big impact when scanning numerous directories/files:

`main`:
---
goos: linux
goarch: amd64
pkg: github.com/chainguard-dev/bincapz/samples
cpu: AMD Ryzen 7 7840U w/ Radeon  780M Graphics     
BenchmarkRun-16    	       1	83237798093 ns/op	761214920 B/op	 4789175 allocs/op
PASS
ok  	github.com/chainguard-dev/bincapz/samples	83.323s

New code:
---
goos: linux
goarch: amd64
pkg: github.com/chainguard-dev/bincapz/samples
cpu: AMD Ryzen 7 7840U w/ Radeon  780M Graphics     
BenchmarkRun-16    	       1	50308758726 ns/op	762749936 B/op	 4789434 allocs/op
PASS
ok  	github.com/chainguard-dev/bincapz/samples	50.402s

We're still hitting single-threaded bottlenecks during certain portions of the code execution so I'm tracking those down.

PR #405, opened 2024-08-15
Use concurrency for path scanning

egibs:
As opposed to #402 which implemented concurrency at the `scanPath` level, this PR implements configurable concurrency at the `path` level, that is, all of the recursively discovered paths within a given `scanPath`. 

By default, the concurrency is limited to `runtime.NumCPU()` and will automatically be set to `1` if the configured concurrency is less than `1`. Similarly, the concurrency can not be set higher than the number of paths being scanned.

I converted `map[any]any` usages over to `go-ordered-map` to ensure that all of our dictionaries maintain their insertion order when iterated over.

Anecdotally, this change has provided a decent increase in speed depending on the concurrency value, number of files, and the size of each file. 

Benchmarking with the default concurrency of `runtime.NumCPU()` takes ~50 seconds:
```
goos: darwin
goarch: arm64
pkg: github.com/chainguard-dev/bincapz/samples
cpu: Apple M3 Pro
BenchmarkRun-12    	       1	48120070292 ns/op	773853464 B/op	 4895873 allocs/op
PASS
ok  	github.com/chainguard-dev/bincapz/samples	49.446s
```

as opposed to ~80 seconds without concurrency (`-j 1`):
```
goos: darwin
goarch: arm64
pkg: github.com/chainguard-dev/bincapz/samples
cpu: Apple M3 Pro
BenchmarkRun-12    	       1	78739845000 ns/op	774942256 B/op	 4896912 allocs/op
PASS
ok  	github.com/chainguard-dev/bincapz/samples	80.366s
```

Anecdotally, scanning `/usr/bin` on my Framework Laptop takes ~60 seconds with concurrency and ~220 seconds without (`-j 1`).


pkg/action/programkind.go
```diff
@@ -80,7 +80,7 @@ func programKind(ctx context.Context, path string) string {
 	headerString := ""
 	n, err := io.ReadFull(f, header[:])
 	if err == nil || errors.Is(err, io.ErrUnexpectedEOF) {
-		kind, err := magic.Lookup(header[:n])
+		kind, err := magic.LookupSync(header[:n])
```
egibs, line 83: I doubt we were seeing much of a benefit from this.
```
LookupSync lookups up the file type based on the provided magic bytes without spawning any additional goroutines. You should provide at least the first 1024 bytes of the file in this slice. A magic.ErrUnknown will be returned if the file type is not known.
```

pkg/action/diff.go
```diff
@@ -25,15 +26,15 @@ func relFileReport(ctx context.Context, c bincapz.Config, fromPath string) (map[
 		return nil, err
 	}
 	fromRelPath := map[string]*bincapz.FileReport{}
-	for _, f := range fromReport.Files {
-		if f.Skipped != "" || f.Error != "" {
+	for files := fromReport.Files.Oldest(); files != nil; files = files.Next() {
```
egibs, line 29: This is the general pattern for `orderedmap`s -- `var.Key` and `var.Value`. Value fields can be accessed via `var.Value.Field`.

pkg/action/diff.go
```diff
@@ -127,7 +128,7 @@ func processDest(ctx context.Context, c bincapz.Config, from, to map[string]*bin
 	for relPath, tr := range to {
 		fr, exists := from[relPath]
 		if !exists {
-			d.Added[relPath] = tr
+			d.Added.Set(relPath, tr)
```
egibs, line 131: This is the pattern for adding a KV pair to an `orderedmap` dictionary.

pkg/action/diff.go
```diff
@@ -246,8 +250,8 @@ func fileMove(ctx context.Context, c bincapz.Config, fr, tr *bincapz.FileReport,
 
 	// Move these into the modified list if the files are not completely different (something like ~0.3)
 	if score > 0.3 {
-		d.Modified[apath] = abs
-		delete(d.Removed, rpath)
-		delete(d.Added, apath)
+		d.Modified.Set(apath, abs)
+		d.Modified.Delete(rpath)
```
egibs, line 254: This is the pattern for deleting a key from an `orderedmap`.

pkg/action/scan.go
```diff
@@ -152,7 +159,7 @@ func recursiveScan(ctx context.Context, c bincapz.Config) (*bincapz.Report, erro
 	logger := clog.FromContext(ctx)
 	logger.Debug("recursive scan", slog.Any("config", c))
 	r := &bincapz.Report{
-		Files: map[string]*bincapz.FileReport{},
+		Files: orderedmap.New[string, *bincapz.FileReport](),
```
egibs, line 162: This is the pattern for initializing a new `orderedmap`.

pkg/action/util.go
```diff
@@ -0,0 +1,24 @@
+// Copyright 2024 Chainguard, Inc.
```
egibs, line 1: These are helper functions for inserting sorted keys into a map.

pkg/action/diff.go
```diff
@@ -153,10 +154,12 @@ func fileDestination(ctx context.Context, c bincapz.Config, fr, tr *bincapz.File
...
 	// are there already modified behaviors for this file?
-	if _, exists := d.Modified[relPath]; !exists {
-		d.Modified[relPath] = abs
+	if _, exists := d.Modified.Get(relPath); !exists {
+		d.Modified.Set(relPath, abs)
 	} else {
-		d.Modified[relPath].Behaviors = append(d.Modified[relPath].Behaviors, abs.Behaviors...)
+		rel, _ := d.Modified.Get(relPath)
+		rel.Behaviors = append(rel.Behaviors, abs.Behaviors...)
+		d.Modified.Set(relPath, rel)
```
egibs, line 162: This was the most difficult portion of migrating to `orderedmap`. I don't think a one-liner was doable but this is more readable.

pkg/action/scan.go
```diff
@@ -191,51 +199,84 @@ func recursiveScan(ctx context.Context, c bincapz.Config) (*bincapz.Report, erro
 		logger.Debug("files found", slog.Any("path count", len(paths)), slog.Any("scanPath", scanPath))
 
 		// path refers to a real local path, not the requested scanPath
+		var wg sync.WaitGroup
```
egibs, line 202: Concurrency changes start here and end on L305 (final construction of the `r.Files` map).

pkg/action/scan.go
```diff
@@ -147,12 +155,35 @@ func errIfHitOrMiss(frs map[string]*bincapz.FileReport, kind string, scanPath st
 	return nil
 }
 
+// structs and types for sorting reports alphabetically by path
+// kv stores the keys and values of a map.
+type kv struct {
```
tstromberg, line 160: This could use a better name. If I saw `actions.kv` somewhere, I would not expect it to hold file reports.

Question: what makes this path different than the one already included in FileReport? If they are the same, we should just sort by that instead.

pkg/action/scan.go
```diff
@@ -147,12 +155,35 @@ func errIfHitOrMiss(frs map[string]*bincapz.FileReport, kind string, scanPath st
...
+	key   string
+	value *bincapz.FileReport
+}
+
+// reports is a slice of paths and their respective file reports.
+type reports []kv
```
tstromberg, line 166: I don't think this type makes the code any clearer, can it be removed? `sort.Slice()` doesn't require special types for sorting.

If we need it for readability, it could definitely use a clearer name, but in my experience, type aliases for slices hurt readability.

tstromberg (changes requested):
Solid improvement, I just have some readability comments regarding  data structures.

pkg/action/scan.go
```diff
@@ -190,8 +222,28 @@ func recursiveScan(ctx context.Context, c bincapz.Config) (*bincapz.Report, erro
 		}
 		logger.Debug("files found", slog.Any("path count", len(paths)), slog.Any("scanPath", scanPath))
 
+		maxConcurrency := c.Concurrency
+		if maxConcurrency < 1 {
+			maxConcurrency = 1
+		} else if maxConcurrency > len(paths) {
+			maxConcurrency = len(paths)
```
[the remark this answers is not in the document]
egibs: It shouldn't. I can remove it and test 👍🏻

pkg/action/scan.go
```diff
@@ -202,38 +254,65 @@ func recursiveScan(ctx context.Context, c bincapz.Config) (*bincapz.Report, erro
...
-			scanPathFindings[path] = fr
-			if !c.OCI {
-				if err := errIfHitOrMiss(map[string]*bincapz.FileReport{path: fr}, "file", path, c.ErrFirstHit, c.ErrFirstMiss); err != nil {
-					logger.Debugf("match short circuit: %s", err)
-					return r, err
+				logger.Debug("processing path", slog.Any("path", path))
+				fr, err := processFile(ctx, c, yrs, path, scanPath, trimPath, logger)
+				if err != nil {
+					mu.Lock()
+					result = append(result, findings{path: path, fr: &bincapz.FileReport{}})
```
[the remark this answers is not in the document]
egibs: I tested this out and got some weird results when running tests where the file reports were merged for two different tests. Managing the mutex(es) isn't ideal but the outcome is deterministic. I'll see if I was doing something wrong with the `sync.Map`, though.

pkg/action/scan.go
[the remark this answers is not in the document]
egibs: Removed in [`c42f418` (#405)](https://github.com/chainguard-dev/bincapz/pull/405/commits/c42f418f0e91112edf176495db48c83051934ba6).

pkg/action/scan.go
```diff
@@ -190,8 +222,28 @@ func recursiveScan(ctx context.Context, c bincapz.Config) (*bincapz.Report, erro
...
+		}
+
 		// path refers to a real local path, not the requested scanPath
+		pc := make(chan string, len(paths))
 		for _, path := range paths {
+			pc <- path
+		}
+		close(pc)
+
+		type findings struct {
```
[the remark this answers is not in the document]
egibs: Removed in [`c42f418` (#405)](https://github.com/chainguard-dev/bincapz/pull/405/commits/c42f418f0e91112edf176495db48c83051934ba6).

pkg/action/scan.go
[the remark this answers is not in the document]
egibs: Yep, I was doing something wrong. I'll push the working code in a bit.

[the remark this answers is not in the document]
egibs: Updated in [`81fca83` (#405)](https://github.com/chainguard-dev/bincapz/pull/405/commits/81fca83d5cfa7ae8641117cf3fcf7f402e7e6dd9).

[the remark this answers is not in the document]
egibs: Follow-up removals/fixes in [`c03d232` (#405)](https://github.com/chainguard-dev/bincapz/pull/405/commits/c03d232eece70d75d3ed81466e19bfbdb8af0423) and [`c4888b3` (#405)](https://github.com/chainguard-dev/bincapz/pull/405/commits/c4888b3b8b236727e6754cb114bb2b0f038e8ff6).

pkg/action/scan.go
egibs: Removed in [`c03d232` (#405)](https://github.com/chainguard-dev/bincapz/pull/405/commits/c03d232eece70d75d3ed81466e19bfbdb8af0423).

pkg/action/scan.go
egibs: Removed in [`c03d232` (#405)](https://github.com/chainguard-dev/bincapz/pull/405/commits/c03d232eece70d75d3ed81466e19bfbdb8af0423).

tstromberg (approved):

tstromberg:
Love it. Thank you!

PR #415, opened 2024-08-19
Remap third-party rule severities instead of ignoring them entirely

egibs:
Closes: https://github.com/chainguard-dev/bincapz/issues/414

Instead of turning off third-party rules that generate critical findings, we can now tweak their respective severities to include the finding without blocking PRs. 

Using `frp` as an example, the finding that was originally a "critical" is now a "high":
CleanShot 2024-08-19 at 10 31 25@2x

We can iterate on this functionality to include our own rules in the future.


egibs:
Chatted with @tstromberg about this a bit. We'll probably leave this behavior as-is and handle "exclusions" on the other side.

PR #421, opened 2024-08-21
Add ability to only show diffs for files with upgraded risks

egibs:
Currently, our diff mode will show all added/removed/modified files. For cases where many files are affected between previous and new paths, this output can be noisy and add little value.

This PR adds the ability to only create file reports for diffs where the new risk score is explicitly higher than the old risk score for a given file.

For example, instead of generating a report for a file that was downgraded from `MEDIUM` to `LOW`, the file will instead be ignored. This behavior defaults to `false` and can be used by adding `--show-upgraded-risks true` when running `bincapz.`


egibs:
I'll probably fold these changes into the CLI overhaul PR I'm working on.

PR #422, opened 2024-08-21
Add test to verify whether bincapz JSON output is valid

egibs:
We're seeing some interesting JSON failures with the `bincapz_check` deployment:
```
"failed to handle workflow run event: failed to scan the package zip file: failed to unmarshal bincapz scan result : unexpected end of JSON input"
```

This was working with `bincapz` < `0.18.0` and may be due to the `orderedmap` changes. This PR adds the following test (obfuscated by the sample archives -- we should at least leave the tests in plaintext, but that's for a future PR):
```go
func TestJSONSerializable(t *testing.T) {
	t.Parallel()
	ctx := slogtest.Context(t)
	clog.FromContext(ctx).With("test", "TestJSON")

	yrs, err := compile.Recursive(ctx, []fs.FS{rules.FS, thirdparty.FS})
	if err != nil {
		t.Fatalf("compile: %v", err)
	}

	fileSystem := os.DirFS(testDataRoot)

	fs.WalkDir(fileSystem, ".", func(path string, _ fs.DirEntry, err error) error {
		if err != nil {
			t.Fatal(err)
		}
		if !strings.HasSuffix(path, ".json") {
			return nil
		}

		name := strings.ReplaceAll(path, ".json", "")
		binPath := filepath.Join(testDataRoot, name)

		// must be a non-test JSON
		if _, err := os.Stat(binPath); err != nil {
			if errors.Is(err, os.ErrNotExist) {
				return nil
			}
			return err
		}

		t.Run(name, func(t *testing.T) {
			var out bytes.Buffer
			render, err := render.New("json", &out)
			if err != nil {
				t.Fatalf("render: %v", err)
			}
			bc := bincapz.Config{
				Concurrency: runtime.NumCPU(),
				IgnoreSelf:  false,
				MinRisk:     1,
				Renderer:    render,
				Rules:       yrs,
				ScanPaths:   []string{binPath},
			}

			tcLogger := clog.FromContext(ctx).With("test", name)
			ctx := clog.WithLogger(ctx, tcLogger)
			res, err := action.Scan(ctx, bc)
			if err != nil {
				t.Fatalf("scan failed: %v", err)
			}

			if err := render.Full(ctx, res); err != nil {
				t.Fatalf("full: %v", err)
			}

			got := out.Bytes()

			// unmarshal the output JSON file into a bincapz report
			scanReport := &bincapz.Report{}
			err = json.Unmarshal(got, scanReport)
			if err != nil {
				t.Errorf("bincapz JSON output is not valid %v", err)
			}
		})
		return nil
	})
}
```

The idea being that we produce a JSON byte slice and then store it in a `bincapz.Report` variable. If the error is not `nil`, then we know that the JSON output cannot be cleanly stored in the variable.

The `bincapz.Report` type uses an `orderedmap` field which used to be a `map[string]*FileReport` type:
```go
type Report struct {
	Files  *orderedmap.OrderedMap[string, *FileReport] `json:",omitempty" yaml:",omitempty"`
	Diff   *DiffReport                                 `json:",omitempty" yaml:",omitempty"`
	Filter string                                      `json:",omitempty" yaml:",omitempty"`
}
```

My working theory is that the `orderedmap` is not compatible with JSON, but I haven't been able to reproduce the JSON error seen in the Cloud Run logs.


egibs:
Not needed. The root cause was that we were unmarshaling an empty report byte slice to JSON.

Issue #426, opened 2024-08-26
Investigate diff performance


egibs:
The diff code path is separate from the usual scanning code path and is much heavier than the latter in the following ways:

1. Each provided path is scanned (this is twice as many file extractions, scans, reports, etc.)
2. The source and destination paths are interrogated to see which files or behaviors appear in one or the other (or both)
3. If files are modified (not added or removed), an O(n^2) loop is used to build a list of reports (though this is not guaranteed to run)
4. A final diff report is returned

Even within the `Diff` code path, steps one, two, and four can run or steps one through four can run.

In cases where we're diffing packages with thousands files, the doubling up can tack on a substantial amount of time (and increase the resource usage overhead). For smaller packages or single files, this is not an issue. We used to only perform diff checks on library files, but there's always the possibility that excluding other file types would ignore critical findings.

egibs:
I'm attempting to grab a profile of a diff between two Spark directories (just `aarch64` and `x86_64` so no version differences).

The memory usage is extremely high but we'll know more once the profiles are rendered.

egibs:
I was able to pare down the memory usage by revisiting the concurrency here: https://github.com/chainguard-dev/bincapz/blob/bdcb6400223228bdf10efa48f9f8c9ed99d8524a/pkg/action/diff.go#L177-L197

When dealing with small numbers of files, this isn't an issue but, again, becomes problematic when looking at thousands or hundreds of thousands of files.

I'm still waiting to see how long a full Spark diff takes with my changes.

PR #429, opened 2024-08-27
Improve diff performance

egibs:
Closes: #426

Diff performance has been a gap as of late. The changes made in #356 were a little too greedy and resulted in calculating the Levenshtein distance and creating reports for files we don't care to diff. 

> I was able to diff Spark and generate a profile for one of the earlier iterations of the code in this PR and the `levenshtein.Calculate` method consuming most of the CPU time:
image

Additionally, the concurrency wasn't implemented correctly which caused a lot of performance issues.

This PR re-adds the filtering of paths and reworks the concurrency used when creating the flatted reports as well as adding concurrency for each of the scans (part of the long execution time was the serial creation of the file reports which will now happen concurrently).

Prior to this change, running a diff on the fully-extracted Spark files would not finish; now the same scan takes ~15 minutes (with two maps consisting of 759,999 and 759,993 elements, respectively):
```
________________________________________________________
Executed in  898.07 secs    fish           external
   usr time   39.35 mins    0.10 millis   39.35 mins
   sys time   39.94 mins    3.66 millis   39.94 mins
```

### DIFF
--- a/pkg/action/diff.go
+++ b/pkg/action/diff.go
@@ -194,8 +194,7 @@ func filterMap(om *orderedmap.OrderedMap[string, *bincapz.FileReport], ps []*reg
 }
 
 // combine iterates over the removed and added channels to create a diff report to store in the combined channel.
-func combine(removed, added <-chan *orderedmap.Pair[string, *bincapz.FileReport], combined chan<- bincapz.CombinedReport, wg *sync.WaitGroup) {
-	defer wg.Done()
+func combine(removed, added <-chan *orderedmap.Pair[string, *bincapz.FileReport], combined chan<- bincapz.CombinedReport) {
 	for r := range removed {
 		for a := range added {
 			score := levenshtein.Match(r.Key, a.Key, levenshtein.NewParams())
@@ -214,13 +213,8 @@ func combine(removed, added <-chan *orderedmap.Pair[string, *bincapz.FileReport]
 }
 
 // combineReports orchestrates the population of the diffs channel with relevant diffReports.
-func combineReports(c bincapz.Config, d *bincapz.DiffReport, combined chan<- bincapz.CombinedReport) {
+func combineReports(d *bincapz.DiffReport, combined chan<- bincapz.CombinedReport) {
 	defer close(combined)
-
-	maxConcurrency := c.Concurrency
-	if maxConcurrency < 1 {
-		maxConcurrency = 1
-	}
 
 	var wg sync.WaitGroup
 
@@ -251,10 +245,7 @@ func combineReports(c bincapz.Config, d *bincapz.DiffReport, combined chan<- bin
 		close(added)
 	}()
 
-	for i := 0; i < maxConcurrency; i++ {
-		wg.Add(1)
-		combine(removed, added, combined, &wg)
-	}
+	combine(removed, added, combined)
 
 	wg.Wait()
 }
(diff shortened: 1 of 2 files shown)


pkg/action/diff.go
```diff
@@ -164,55 +183,98 @@ func fileDestination(ctx context.Context, c bincapz.Config, fr, tr *bincapz.File
 	}
 }
 
-type diffReports struct {
-	Added     string
-	AddedFR   *bincapz.FileReport
-	Removed   string
-	RemovedFR *bincapz.FileReport
+// filterMap filters orderedmap pairs by checking for matches against a slice of compiled regular expression patterns.
+func filterMap(om *orderedmap.OrderedMap[string, *bincapz.FileReport], ps []*regexp.Regexp, c chan<- *orderedmap.Pair[string, *bincapz.FileReport], g *errgroup.Group) {
```
egibs, line 187: This is where we really pare down what we need to iterate through when calculating the Levenshtein distance and merging the two reports.

pkg/action/diff.go
```diff
@@ -164,55 +183,101 @@ func fileDestination(ctx context.Context, c bincapz.Config, fr, tr *bincapz.File
...
-	go func() {
-		wg.Wait()
-		close(diffs)
-	}()
+	var g errgroup.Group
+	g.SetLimit(maxConcurrency)
+
+	// Patterns we care about when handling diffs
+	patterns := []string{
+		`^[\w.-]+\.(so|a|dylib|dll|lib)$`,
```
tstromberg, line 239: I believe we only need `.so` here - the other extension styles don't typically changed based on the version.

tstromberg (changes requested):

pkg/action/diff.go
egibs: Updated in [`e9fb45e` (#429)](https://github.com/chainguard-dev/bincapz/pull/429/commits/e9fb45e8d59d0604d032540b37708bbe892b5df2#diff-296f8d8ff6837443670f31f7e7dd79d8908c01dc9059790f2ab6d1c78cafacdbR232).

pkg/action/diff.go
```diff
@@ -164,55 +181,93 @@ func fileDestination(ctx context.Context, c bincapz.Config, fr, tr *bincapz.File
...
-		reports = append(reports, diff)
-	}
-	return reports
+	wg.Add(1)
+	go func() {
+		filterMap(d.Added, ps, added, &wg)
+		close(added)
+	}()
+
+	combine(removed, added, combined)
```
egibs, line 253: I removed concurrency from this since we're iterating over such a small number of items compared to before when nothing was filtered.

pkg/action/diff.go
```diff
@@ -164,55 +181,93 @@ func fileDestination(ctx context.Context, c bincapz.Config, fr, tr *bincapz.File
...
+	ps := make([]*regexp.Regexp, len(patterns))
+	for i, pattern := range patterns {
+		ps[i] = regexp.MustCompile(pattern)
 	}
 
+	// Build two channels with filtered paths to iterate through in the worker pool
+	removed := make(chan *orderedmap.Pair[string, *bincapz.FileReport], d.Removed.Len())
+	added := make(chan *orderedmap.Pair[string, *bincapz.FileReport], d.Added.Len())
+
+	wg.Add(1)
```
egibs, line 241: Similar to scanning each of the paths concurrently, we'll also filter the resulting removed/added maps concurrently.

tstromberg (approved):

pkg/action/diff.go
egibs: Prior to me opening this PR, I had added a worker pool pattern to handle the maps in chunks to reduce memory usage, but since we're filtering the maps there's not a need for that any longer.

Issue #428, opened 2024-08-27
Replace OCI test image with crane extraction


egibs:
We shouldn't be pulling the image when running the OCI test. Instead, we should use crane to create a persistent `.tar.gz` of the image that we store next to the code.

PR #438, opened 2024-09-04
Replace live OCI image pull with crane export

egibs:
Closes: #428

This PR removes our dependency on a non-deterministic image source with a static tarball created via `crane export`. 

While this won't engage the `if c.OCI` logic, this is what that functionality does anyway: https://github.com/chainguard-dev/bincapz/blob/d3d7a6bb7fc13a4dfeeb26711c5617e4c4496a8d/pkg/action/oci.go#L15-L40

The scan results are the same (sans the base path which has been updated).

For future reference, the command I ran to create the tarball was:
```
cd pkg/action/testdata; crane export cgr.dev/chainguard/static:latest static.tar
```


tstromberg:
Would this test still work if the sample was compressed to a .tar.xz? If so, let's do that.

egibs:
> Would this test still work if the sample was compressed to a .tar.xz? If so, let's do that.

It does -- updated in [`7bf82eb` (#438)](https://github.com/chainguard-dev/bincapz/pull/438/commits/7bf82ebb568e0a56d2b9f6b264a3d47e6dc4743f).

Makefile
```diff
@@ -123,3 +123,11 @@ clone-samples:
 		tar -xJvf samples/linux/clean/$$file -C samples/linux/clean; \
 	done
 	tar -xJvf samples/macOS/clean/bincapz.xz -C samples/macOS/clean
+
+.PHONY: install-crane
+install-crane:
+	go install github.com/google/go-containerregistry/cmd/crane@latest
```
tstromberg, line 129: make shouldn't modify files outside of this directory. I suggest setting GOBIN and pinning this to a specific version for consistent output. This will also allow you to cache the binary without reinstalling it every run. Example:

https://github.com/chainguard-dev/osquery-defense-kit/blob/df577d4f1c08053a4a3bc53a5461736662fb3c82/Makefile#L6

egibs: Makes sense 👍🏻 

Updated in [`d220c8f` (#438)](https://github.com/chainguard-dev/bincapz/pull/438/commits/d220c8fe0e66e1c1e5fd5085d3868900c3622f2a).

Makefile
```diff
@@ -123,3 +123,13 @@ clone-samples:
 		tar -xJvf samples/linux/clean/$$file -C samples/linux/clean; \
 	done
 	tar -xJvf samples/macOS/clean/bincapz.xz -C samples/macOS/clean
+
+CRANE_VERSION=v0.20.2
+.PHONY: install-crane
```
tstromberg, line 128: Phony targets aren't cached. Instead of a phony target, can you make this a filename-based make target (including the version)?

egibs: Updated in [`ef6376b` (#438)](https://github.com/chainguard-dev/bincapz/pull/438/commits/ef6376b71e5c4be65ca05a8525231cc2c4fe6f2e). I fixed the OCI test refresh to pull from the `.xz` archive as well (and cleaned up the path to remove `../pkg/action/`).

Makefile
```diff
@@ -123,3 +123,13 @@ clone-samples:
 		tar -xJvf samples/linux/clean/$$file -C samples/linux/clean; \
 	done
 	tar -xJvf samples/macOS/clean/bincapz.xz -C samples/macOS/clean
+
+ARCH ?= $(shell uname -m)
+CRANE_VERSION=v0.20.2
+install-crane-$(ARCH)-$(CRANE_VERSION):
```
tstromberg, line 129: To make caching work properly, the Makefile rule should be the output filename that that rule.

For example, `./out/crane-<whatever>`

Makefile
```diff
@@ -123,3 +123,13 @@ clone-samples:
...
+	mkdir -p out
+	GOBIN=$(CURDIR)/out go install github.com/google/go-containerregistry/cmd/crane@$(CRANE_VERSION)
+
+.PHONY: extract-image
+extract-image: install-crane-$(ARCH)-$(CRANE_VERSION)
```
tstromberg, line 134: "extract" doesn't feel like the right word for this action - maybe "dump" or "export"?

tstromberg (changes requested):

Makefile
egibs: Updated in [`f59bdef` (#438)](https://github.com/chainguard-dev/bincapz/pull/438/commits/f59bdef7cce505502d1550ee5295604d457b63e1).

Makefile
egibs: Fixed in [`f5098de` (#438)](https://github.com/chainguard-dev/bincapz/pull/438/commits/f5098deb8d74b545f6cd590d1dd072367054c1e9).

tstromberg (approved):

PR #450, opened 2024-09-15
refresh-sample-testdata refactor

tstromberg:
- Improve parallelism
- Fix Makefile dependency
- Remove extraneous cp
- Make it easier to select alternative repos:

```
make SAMPLES_REPO=tstromberg/bincapz-samples SAMPLES_COMMIT=7e2d002 refresh-sample-testdata
```

On my MacBook Air, this PR decreases the makefile rule execution time by about 50%.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,17 +47,17 @@ fix: $(FIXERS)
 
 # END: lint-install ../bincapz
 
-SAMPLES_REPO=https://github.com/chainguard-dev/bincapz-samples.git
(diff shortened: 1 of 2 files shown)


tstromberg:
Sorry, had to disable commit signing so I could fix this on the plane :)

egibs (approved):
Thanks for knocking this out. I unintentionally neglected this portion when doing all of the other work to split out the samples.

PR #468, opened 2024-09-24
Add --processes flag to scan active process commands

egibs:
This PR adds the ability to scan the commands of running processes if the command is a valid, absolute path. 

`go-yara` provides a `ScanProc` method; however, it was proving difficult to use (at least on MacOS where SIP may be a contributing factor), so this implementation focuses on the binaries of running processes which is actually easier to fit into our current scan pattern (loading file descriptors).

To support the scanning of several hundred process paths, I had to track down and address a bug with the `scanPathFindings` `sync.Map`. We were not removing keys as they were used so we'd endlessly iterate through them as the Map elements grew. This wasn't originally apparent because we would provide a single scan path (even if it was a directory).

Finally, I also fixed the brief terminal output and made it a bit easier to read when accounting for multiple lines:
CleanShot 2024-09-24 at 17 48 09@2x


pkg/action/process.go
```diff
@@ -0,0 +1,101 @@
...
+			}
+		}
+	}
+
+	return mapToSlice(uniquePaths), nil
+}
+
+// getWindowsProcessPaths is a Windows-focused function to retrieve PIDs and their respective commands.
+func getWindowsProcessPaths() ([]Process, error) {
+	cmd := exec.Command("wmic", "process", "get", "ProcessId,ExecutablePath")
```
egibs, line 59: For reference: https://superuser.com/a/1747592

I can potentially test this out later.

pkg/action/scan.go
```diff
@@ -324,7 +324,7 @@ func recursiveScan(ctx context.Context, c malcontent.Config) (*malcontent.Report
 
 		// Add the sorted paths and file reports to the parent report and render the results
 		for _, k := range pathKeys {
-			finding, ok := scanPathFindings.Load(k)
+			finding, ok := scanPathFindings.LoadAndDelete(k)
```
egibs, line 327: It took a while to track this down but luckily this is all that was needed to avoid needing to loop through each scan path one at a time in the base command code in `malcontent.go`.

malcontent.go
```diff
@@ -363,9 +371,24 @@ func main() {
 					switch {
 					case c.String("image") != "":
 						mc.OCI = true
-					case c.String("image") == "":
+					case c.String("image") == "" && !c.Bool("processes"):
```
egibs, line 374: This condition was needed to prevent a no-op exit if no scan paths were provided to the command.

egibs:
Updated my fork to the wrong remote URL. Closing this out and remaking the PR.

pkg/action/process.go
```diff
@@ -0,0 +1,101 @@
...
+	"strings"
+)
+
+type Process struct {
+	PID  int
+	Path string
+}
+
+// GetAllProcessPaths is an exported function that retrieves the process commands.
+func GetAllProcessPaths() ([]Process, error) {
```
tstromberg, line 18: It may be overkill for this, but https://github.com/shirou/gopsutil is a well-regarded API for grabbing a process table in Go.

egibs: I would grealy enjoy a more idiomatic way of doing this 🙂

PR #477, opened 2024-10-02
malcontent branding: rewrite README, new go install target

tstromberg:
* New `go install` target to preserve `mal` binary name: `go install github.com/chainguard-dev/malcontent/cmd/mal@latest`
* Refactor README so that it's shorter and captures the reality of the tool
* New DEVELOPMENT.md doc


DEVELOPMENT.md
```diff
@@ -0,0 +1,79 @@
...
+yara -s -w rules/combo/dropper/shell.yara <path>
+```
+
+## Running tests
+
+```make test```
+
+## Writing rule tests
+
+Not every rule needs a test, but tests do unsure that a rules behavior stays consistent:
```
egibs, line 26:
```suggestion
Not every rule needs a test, but tests do ensure that a rules behavior stays consistent:
```

egibs (approved):

PR #479, opened 2024-10-02
Small archive extraction fixes; support bzip2 archives

egibs:
We noticed a few issues around extracting archives and I also noticed that we never added bzip2 support.

This PR:
- fixes a bug where we wouldn't seek to the file origin for `.apk`, `.gz`, and `.tar.gz` files (that statement never ran `tf.Seek(0, io.SeekStart)`) by moving the seek code out of the switch statements
- fixes a bug where we didn't check for `io.ErrUnexpectedEOF` when iterating through the `tar` reader
- adds `bzip2` archive support

### DIFF
--- a/pkg/action/archive.go
+++ b/pkg/action/archive.go
@@ -3,6 +3,7 @@ package action
 import (
 	"archive/tar"
 	"archive/zip"
+	"compress/bzip2"
 	"compress/gzip"
 	"context"
 	"errors"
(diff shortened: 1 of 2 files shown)


pkg/action/archive.go
```diff
@@ -58,8 +60,13 @@ func extractTar(ctx context.Context, d string, f string) error {
 			return fmt.Errorf("failed to create xz reader: %w", err)
 		}
 		tr = tar.NewReader(xzStream)
+	case strings.Contains(filename, ".bz2") || strings.Contains(filename, ".bzip2"):
+		if err != nil {
+			return fmt.Errorf("failed to seek to start: %w", err)
+		}
+		br := bzip2.NewReader(tf)
```
egibs, line 67: It's weird that `bzip2.NewReader` only returns an `io.Reader` instead of `(io.Reader, err)`. It would be nice to do some error checking like we do with `gzip.NewReader`. 😞

tstromberg (approved):

PR #473, opened 2024-09-30
Add support for rule overrides; ignore "false_positive" and "ignore" tags by default

egibs:
Closes #471

We've wanted to support rule overrides for a while. This PR attempts to do so in a basic way by storing matched rules in a map (both the original and the override should match) and then dropping the original rule if an override exists. For now, we'll allow for severity overrides and then store the combined descriptions of the two rules for later display.

For example,

This YARA:
```yara
rule send {
	meta:
		description = "send a message to a socket"
		ref = "https://linux.die.net/man/2/send"
		syscall = "send"
	strings:
		$send = "send" fullword
		$socket = "socket" fullword
	condition:
		all of them
}

rule send_override {
    meta:
        description = "this overrides the send rule"
        ref = "https://linux.die.net/man/2/send"
        syscall = "send"
    strings:
        $send = "send" fullword
        $socket = "socket" fullword
    condition:
        all of them
}
```

Results in this output:
```
$ go run . analyze out/samples-84a2015439ad84d3005fa57d06d7a925acb9f2a3/javascript/clean/connection.js
out/samples-84a2015439ad84d3005fa57d06d7a925acb9f2a3/javascript/clean/connection.js [⚠️ MEDIUM]
---------------------------------------------------------------------------------------------------------------------------------
RISK  KEY                         DESCRIPTION                                        EVIDENCE
---------------------------------------------------------------------------------------------------------------------------------
LOW   encoding/base64             Supports base64 encoded strings                    base64
LOW   fd/write                    writes to a file handle                            outgoingMessageStream.write(message)
LOW   net/dns                     Uses DNS (Domain Name Service)                     require("dns")
LOW   net/socket/send             this overrides the send rule, send a message to a  send
                                  socket                                             socket
LOW   ref/site/url                contains embedded HTTPS URLs                       https://github.com/pekim/tedious/issues/24
LOW   ref/words/password          references a 'password'                            UsernamePasswordCredential
                                                                                     directory-password
MED   data/embedded/base64/terms  Contains base64 content                            FkZHJlc3::$address
                                                                                     RpcmVjdG9ye::$directory
                                                                                     YWRkcmVzc::$address
                                                                                     ZGlyZWN0b3J5::$directory
                                                                                     aHRtb::$html
                                                                                     hZGRyZXNz::$address
                                                                                     kaXJlY3Rvcn::$directory
MED   data/embedded/base64/url    Contains base64 url                                aHR0cDovL::$http
                                                                                     h0dHA6Ly::$http
[...]

Issue #471, opened 2024-09-26
Provide override rules that lower the priority of matches


tstromberg:
At the moment, malcontent handles false-positives by hiding them. Often times though, we want the result to still show up in `analyze`, but at a lower priority. For example, we have this rule:

```
rule bash_history : high {
  meta:
    description = "access .bash_history file"
  strings:
    $ref = ".bash_history" fullword
  condition:
    all of them
}
```

However, `/bin/bash` will trigger this, which it's correct to do, but malcontent shouldn't categorize it as `high` as a known-good program. My naive first thought is adding support for override rules that change the priority of a matching rule:

```
rule bash_history_override_bash : low {
strings:
  $ref = "BASH_VERSION" fullword
  $ref2 = "BASH_REMATCH" fullword
condition:
  bash_history and filesize < 5MB and all of them  
}
```

The result would then be a low-priority result that says something like `access .bash_history file (bash)`: combining the original description and the override suffix. Nothing stops us from using this technique to explicitly raise the priority of a match, though we implicitly do that already. This could improve upon the output there.

tstromberg:
This is related to #414

egibs:
#473 contains an attempt to do this.

The rule name has to follow the `original_rule_override` pattern but there may be a way to also allow variants of overrides (but this will be consistent).

If an override for a rule exists, we'll ignore the original rule and use the override's severity with the new description being the original rule's description appended to the override rule's description.

tstromberg:
@egibs brought up some good points about handling 3rd party rules to me in a conversation. This also had me thinking that this implementation is inefficient:

- Every override effectively doubles rule computation time
- If a piece of 3rd party software causes multiple false positives, we have to re-implement the matching logic each time

Here's a twist on the idea that addresses both:

```
rule bash_history : high {
  meta:
    description = "access .bash_history file"
    decrease_if = "bash"
    drop_if = ""
  strings:
    $ref = ".bash_history" fullword
  condition:
    all of them
}

rule its_bash : override {
  meta:
    bash_history = "medium"
    third_party_malware_rule = "false_positive"
  strings:
    $ref = "BASH_VERSION" fullword
    $ref2 = "BASH_REMATCH" fullword
  condition:
    filesize < 5MB and all of them  
}
```

This shows an example of a single override rule that modifies the priority of two rules: a third-party rule and a local rule. 

[...]

PR #481, opened 2024-10-03
Allow for rule severity overrides; add default ignore tags

egibs:
#473 got a little messy after I refactored everything (+ merge conflicts).

This PR implements the suggestions in #471; namely, the ability to override a rule's severity via metadata.

To do this, we generate a map of rule matches, check if a metadata key exists in that map, and then iterate through the file's behaviors to update the matching rule's severity with the value of the metadata field.

We've also wanted to ignore rules based on tags, so I added default ignores for `false_positive` or `ignore` tags (and add them to the `ignoreTags` slice if they're omitted via the CLI).

### DIFF
--- a/pkg/report/report.go
+++ b/pkg/report/report.go
@@ -344,9 +344,18 @@ func Generate(ctx context.Context, path string, mrs yara.MatchRules, c malconten
 	}
 
 	for _, m := range mrs {
+		override := false
 		if all(m.Rule == NAME, ignoreSelf) {
 			ignoreMalcontent = true
 		}
+
+		for _, t := range m.Tags {
+			if t == "override" {
+				override = true
+				break
+			}
+		}
+
 		risk = behaviorRisk(m.Namespace, m.Rule, m.Tags)
 		if risk > overallRiskScore {
 			overallRiskScore = risk
@@ -392,8 +401,9 @@ func Generate(ctx context.Context, path string, mrs yara.MatchRules, c malconten
 			}
 
 			// If we find a match in the map for the metadata key, that's the rule to override
+			// Ensure that the override tag is present on the override rule
 			var overrideRule string
-			if match, exists := mrsMap[k]; exists {
+			if match, exists := mrsMap[k]; exists && override {
 				overrideRule = match.Rule
 			}
 
@@ -439,17 +449,26 @@ func Generate(ctx context.Context, path string, mrs yara.MatchRules, c malconten
 				syscalls = append(syscalls, sy...)
 			case "cap":
 				caps = append(caps, v)
-			// If we find a rule to override, set that rule's risk score and level
+			// If we find a rule to override, pull in that rule's configuration and update the severity
 			case overrideRule:
 				var overrideSev int
 				if sev, ok := Levels[v]; ok {
 					overrideSev = sev
 				}
-				// Find the behavior for the overridden rule
-				for _, b := range fr.Behaviors {
-					if b.RuleName == overrideRule {
+				// Find the behavior for the overridden (original) rule
+				// Store its behavior in the current behavior and remove the original behavior
+				for i, ob := range fr.Behaviors {
+					if ob.RuleName == overrideRule {
+						if b.Description != "" {
+							b.Description = fmt.Sprintf("%s, %s", b.Description, ob.Description)
+						} else {
+							b.Description = ob.Description
+						}
 						b.RiskScore = overrideSev
 						b.RiskLevel = RiskLevels[overrideSev]
+
+						// Remove the original rule from the behaviors slice
+						fr.Behaviors = slices.Delete(fr.Behaviors, i, i+1)
 					}
 				}
 			}
--- a/pkg/report/report.go
+++ b/pkg/report/report.go
@@ -295,7 +295,7 @@ func mungeDescription(s string) string {
 	return s
 }
 
-//nolint:cyclop,gocognit // ignore complexity of 44, 103
+//nolint:cyclop // ignore complexity of 44
 func Generate(ctx context.Context, path string, mrs yara.MatchRules, c malcontent.Config, expath string) (malcontent.FileReport, error) {
 	ignoreTags := c.IgnoreTags
 	minScore := c.MinRisk
@@ -461,7 +461,7 @@ func Generate(ctx context.Context, path string, mrs yara.MatchRules, c malconten
 					if ob.RuleName == overrideRule {
 						b = ob
 						b.RuleName = m.Rule
-						b.Description = fmt.Sprintf("%s, overridden by %s", b.Description, m.Rule)
+						b.Description = fmt.Sprintf("%s, [%s]", b.Description, m.Rule)
 						b.RiskScore = overrideSev
 						b.RiskLevel = RiskLevels[overrideSev]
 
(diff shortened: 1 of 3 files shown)


pkg/report/report.go
```diff
@@ -383,6 +391,12 @@ func Generate(ctx context.Context, path string, mrs yara.MatchRules, c malconten
 				continue
 			}
 
+			// If we find a match in the map for the metadata key, that's the rule to override
+			var overrideRule string
+			if match, exists := mrsMap[k]; exists {
```
tstromberg, line 396: This should check that the current rule has an override tag, otherwise we are bound to run into weird edge cases where a rule has a metadata field matching another rule name.

tstromberg (changes requested):

pkg/report/report.go
```diff
@@ -439,17 +449,26 @@ func Generate(ctx context.Context, path string, mrs yara.MatchRules, c malconten
...
 				if sev, ok := Levels[v]; ok {
 					overrideSev = sev
 				}
-				// Find the behavior for the overridden rule
-				for _, b := range fr.Behaviors {
-					if b.RuleName == overrideRule {
+				// Find the behavior for the overridden (original) rule
+				// Store its behavior in the current behavior and remove the original behavior
+				for i, ob := range fr.Behaviors {
+					if ob.RuleName == overrideRule {
```
egibs, line 461: @tstromberg -- let me know if you think we should pull in any of the original rule's fields (author, URL, etc.).

pkg/report/report.go
egibs: Added in [`2fa8a07` (#481)](https://github.com/chainguard-dev/malcontent/pull/481/commits/2fa8a07151d287575f18ef4313e190a93c5b51a8).

egibs:
> This should also ignore drop or otherwise hide results from rules with an override tag.

Added in [`2fa8a07` (#481)](https://github.com/chainguard-dev/malcontent/pull/481/commits/2fa8a07151d287575f18ef4313e190a93c5b51a8).

I verified that this worked by running your example bash rules:
```yara
// unchanged bash_history rule
rule bash_history : high {
  meta:
    description = "access .bash_history file"
  strings:
    $ref = ".bash_history" fullword
  condition:
    all of them
}

rule yes_it_is_really_bash : override {
  meta:
    bash_history = "medium"
    third_party_malware_rule = "false_positive"
  strings:
    $ref = "BASH_VERSION" fullword
    $ref2 = "BASH_REMATCH" fullword
  condition:
    filesize < 5MB and all of them  
}
```
⬇️
```
/usr/bin/bash [🔥 HIGH]
---------------------------------------------------------------------------------------------------------------------
RISK  KEY                      DESCRIPTION                                         EVIDENCE                          
---------------------------------------------------------------------------------------------------------------------
LOW   env/HOME                 Looks up the HOME directory for the current user    HOME                              
                                                                                   getenv                            
LOW   env/LANG                 Looks up language of current user                   LANG                              
                                                                                   getenv                            
LOW   env/SHELL                path to active shell                                SHELL                             
LOW   env/TERM                 Look up or override terminal settings               TERM                              
LOW   env/TMPDIR               TMPDIR                                              TMPDIR                            
                                                                                   getenv                            
LOW   exec/program/background  wait for process to exit                            waitpid                           
LOW   fs/file/delete           deletes files                                       unlink                            
LOW   fs/link/read             read value of a symbolic link                       readlink                          
LOW   fs/symlink/resolve       resolves symbolic links                             realpath                          
LOW   fs/tempdir               looks up location of temp directory                 TMPDIR                            
LOW   fs/tempdir/create        creates temporary directory                         mkdtemp                           
LOW   fs/tempfile/create       Uses mktemp to create temporary files               mktemp                            
                                                                                   temp file                         
LOW   kernel/hostname/get      get computer host name                              gethostname                       
LOW   net/hostport/parse       Network address and service translation             freeaddrinfo                      
                                                                                   getaddrinfo                       
LOW   net/socket/listen        listen on a socket                                  accept                            
                                                                                   socket                            
LOW   net/socket/peer/address  get peer address of connected socket                getpeername                       
LOW   process/chdir            changes working directory                           cd which change the               
LOW   process/executable_path  sets a custom PATH                                  /bin:/usr/                        
                                                                                   /sbin:/bin                        
                                                                                   /usr/bin:/sbin                    
                                                                                   PATH                              
LOW   process/groupid/set      set real, effective, and saved group ID of process  setresgid                         
LOW   process/parent_pid/get   gets parent process ID                              getppid                           
LOW   process/userid/set       set real, effective, and saved user ID of process   setresuid                         
LOW   ref/path/etc             path reference within /etc                          /etc/bash.bashrc                  
                                                                                   /etc/hosts                        
                                                                                   /etc/inputrc                      
                                                                                   /etc/profile                      
LOW   ref/path/var             path reference within /var                          /var/tmp                          
LOW   ref/site/url             contains embedded HTTP URLs                         http://gnu.org/licenses/gpl.html  
                                                                                   http://www.gnu.org/gethelp/       
                                                                                   http://www.gnu.org/software/bash  
MED   dylib/symbol/address     get the address of a symbol                         dlsym                             
MED   exec/program             executes external programs                          execve                            
MED   fs/permission/modify     modifies file permissions                           chmod                             
MED   group/lookup             get entry from group database                       endgrent                          
                                                                                   getgrent                          
                                                                                   setgrent                          
MED   net/ssh                  Uses SSH (secure shell) service                     SSH                               
MED   ref/path/etc/hosts       references /etc/hosts                               /etc/hosts                        
MED   ref/path/usr/local       path reference within /usr/local/bin                /usr/local/bin                    
MED   secrets/bash_history     access .bash_history file                           BASH_REMATCH                      
                                                                                   BASH_VERSION                      
MED   shell/exec               executes shell                                      /bin/sh                           
MED   tty/pathname             returns the pathname of a terminal device           ttyname                           
HIGH  exec/shell_command       execute a shell command                             shell_execve                      
---------------------------------------------------------------------------------------------------------------------
```
⬇️
```
MED   secrets/bash_history     access .bash_history file                           BASH_REMATCH                      
                                                                                   BASH_VERSION
```

pkg/report/report.go
tstromberg: My suggestion: preserve everything in the original rule; just override the severity and add a suffix to the description.

egibs: Updated in: [`bbe0701` (#481)](https://github.com/chainguard-dev/malcontent/pull/481/commits/bbe07015e121cdf0d4db52791800a11c645b5622)

Example of the description suffix:
```
MED   secrets/bash_history     access .bash_history file, overridden by            .bash_history                     
                               yes_it_is_really_bash                                            
```

tstromberg:
Were you able to test this with the example to confirm behavior?

One issue to eventually aff5sss is that we currently allow duplicate named rules across files. I don't think this will break anything though with those cases unless we are overriding them.

We can restrict duplicate named rules in another PR, I'm guessing we probably have 20 or so.

tstromberg:
Amazing work. I can't wait to use this! 

My suggestion for the description suffix: just add something short like " (override_rule)" or " [override_rule]".

For example, we would probably name the override something short like "bash".

tstromberg (approved):

PR #487, opened 2024-10-04
Better handling of overrides after all fr.Behaviors are added

### LINKED ISSUE #486: Rule overrides can fail if the override is iterated over before the original
I noted the possibility of this in #473 ([comment](https://github.com/chainguard-dev/malcontent/pull/473#discussion_r1785299311)), but I encountered the edge case where the rule that is being overridden has not been iterated over, so the override never satisfies the `ob.RuleName == overrideRule` condition.

I'm working through the logic to will resolve this but I wanted to post the Issue for visibility.

egibs:
Closes #486

We hit the edge case where an override rule was being evaluated before the original rule, so we never found it in the slice of behaviors.

This PR handles overrides as a post-processing step, rather than during the `mrs` loop.

This PR also introduces our first Critical -> High override for a third-party (!) rule:
```
HIGH  override/plugx                  detect PlugX in memory, [malware_PlugX_config], by   $v2b                                                                                                                     
                                      JPCERT/CC Incident Response Group                    $v2f                                                                                                                     
                                                                                           $v2g                                                                                                                     
                                                                                           $v2h                                                                                                                     
                                                                                           �DDDD                                                                                                                    
                                                                                           �DDDD                                                                                                                    
                                                                                           �3333                                                                                                                    
                                                                                           �DDDD  
```

Because of this rule, I was able to see that the original override logic was insufficient. The new approach is more resilient and we should successfully apply all overrides now.


pkg/report/report.go
```diff
@@ -386,12 +401,6 @@ func Generate(ctx context.Context, path string, mrs yara.MatchRules, c malconten
 		k := ""
 		v := ""
 
-		// Store match rules in a map for future override operations
-		mrsMap := make(map[string]yara.MatchRule)
```
egibs, line 390: This was in the wrong spot as well.

pkg/report/report.go
```diff
@@ -521,6 +517,59 @@ func Generate(ctx context.Context, path string, mrs yara.MatchRules, c malconten
...
+	for _, vo := range validOverrides {
+		original, originalExists := originalMap[vo.Original]
+		override, overrideExists := overrideMap[vo.Override]
+
+		// If the original and override rules exist, 
+		// update the override rule with the correct severity and description from the original
+		if originalExists && overrideExists {
+			for _, b := range fr.Behaviors {
+				if b.RuleName == override.RuleName {
+					b.RuleAuthor = original.RuleAuthor
```
egibs, line 560: I was getting incorrect results if I did `b = original`, so I was more prescriptive with the fields here. I think it adds helpful context anyway.

rules/override/plugx.yara
```diff
@@ -0,0 +1,21 @@
+rule plugx : override {
+    meta:
+        malware_PlugX_config = "high"
```
egibs, line 3: 🎉

egibs:
Currently working through a discrepancy between `analyze` and `scan`. The former works, but the latter seems to ignore the override completely (i.e., `validOverrides` isn't even populated).

tstromberg (changes requested):
Could you make it so that overridden rules use the original name rather than the override name?

rules/override/plugx.yara
tstromberg: this override could use a description; I can't figure out what the override is for or what these string values are.

egibs:
> Could you make it so that overridden rules use the original name rather than the override name?

Updated in [`01ddf7d` (#487)](https://github.com/chainguard-dev/malcontent/pull/487/commits/01ddf7db41d2729b0df98b790783dfb16e29b5b9).

rules/override/shellcode.yara
```diff
@@ -0,0 +1,122 @@
+rule shellcode_hash : override {
+    meta:
+        malware_shellcode_hash = "high"
+
+    strings:
```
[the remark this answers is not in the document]
tstromberg: We should invert how this override functions: it should simply detect the appropriate  Pulumi binary rather than repeat the logic of the offending rule.

Here's what I would do in your shoes:

- Rename this file to "pulumi.yara"
- Rename the rule to the binary you are overriding, like `pulumictl` (if it's not obviously pulumi, maybe "pulumi_<binary>".
- Add 1-2 long strings you expect to show up in the binary, such as "PULUMI_HOME" or "How did Pulumi come up with such a weird name?"

rules/override/shellcode.yara
```diff
@@ -0,0 +1,122 @@
+rule shellcode_hash : override {
+    meta:
+        malware_shellcode_hash = "high"
+
```
tstromberg, line 4: Add a description: for example: are we raising or lowering the severity, and why?

rules/override/shellcode.yara
```diff
@@ -0,0 +1,122 @@
+rule shellcode_hash : override {
+    meta:
+        malware_shellcode_hash = "high"
```
tstromberg, line 3: I couldn't find a rule with this name:

https://github.com/search?q=repo%3Achainguard-dev%2Fmalcontent+malware_shellcode_hash&type=code

tstromberg (changes requested):

tstromberg:
Ideal: instead of a generic `overrides/` folder within rules, maybe a `known_good/` folder specifically geared toward addressing false positives?

I figure that overrides that add severity should probably live elsewhere?

rules/override/shellcode.yara
egibs: It exists in two places:
- https://github.com/chainguard-dev/malcontent/blob/0acb2e0e98915c8fee1d6723dfbf1a29d861ed5c/third_party/yara/JPCERT/shellcode.yara#L1
- https://github.com/chainguard-dev/malcontent/blob/0acb2e0e98915c8fee1d6723dfbf1a29d861ed5c/third_party/yara/JPCERT/other/shellcode.yara#L1

The override works across both, though:
```
HIGH  3P/JPCERT/shellcode/hash       detect shellcode api hash value, by JPCERT/CC Incident  $or60hAddShl1Hash32_GetProcAddress
                                     Response Group                                          $or60hAddShl1Hash32_LoadLibraryA
```

rules/override/shellcode.yara
egibs: Will do. This was mostly just to test that everything was working. I can leave these out of this PR and add them in the next false positive fix PR with the correct details.

rules/override/shellcode.yara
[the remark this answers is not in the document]
egibs: Makes sense. Your Bash example comes to mind and that's a good way to go about it.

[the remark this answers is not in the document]
egibs: Flipped the logic in [`a4c70f4` (#487)](https://github.com/chainguard-dev/malcontent/pull/487/commits/a4c70f42836206fc80d72f275130f309a8a56037).

pkg/report/report.go
egibs: Flipped the logic in [`a4c70f4` (#487)](https://github.com/chainguard-dev/malcontent/pull/487/commits/a4c70f42836206fc80d72f275130f309a8a56037).

rules/override/plugx.yara
egibs: Yep, just a test. Removed in [`a4c70f4` (#487)](https://github.com/chainguard-dev/malcontent/pull/487/commits/a4c70f42836206fc80d72f275130f309a8a56037).

I'll work on a better rule in the next false-positive PR.

pkg/report/report.go
```diff
@@ -521,6 +517,72 @@ func Generate(ctx context.Context, path string, mrs yara.MatchRules, c malconten
 		// TODO: If we match multiple rules within a single namespace, merge matchstrings
 	}
```
tstromberg, line 519: Can some of the override handling code move to a new function? This one is getting overly large and complicated.

pkg/report/report.go
```diff
@@ -521,6 +517,72 @@ func Generate(ctx context.Context, path string, mrs yara.MatchRules, c malconten
...
+				overrideMap[vo.Override] = b
+				isOverride = true
+				break
+			}
+		}
+
+		// If this is the override rule, delete it from the behavior slice
+		// Otherwise, treat the rule as a possible original (vo.Original makes this an O(1) lookup)
+		if isOverride {
+			fr.Behaviors = slices.Delete(fr.Behaviors, i, i+1)
```
tstromberg, line 544: Instead of deleting overrides from Behaviors, would the logic be simpler if overrides were stored elsewhere instead? For example, `fr.Overrides` or a different variable altogether.

By the time `fr.Behaviors` is populated, we already know if the rule has an `override` tag.

pkg/report/report.go
```diff
@@ -521,6 +517,72 @@ func Generate(ctx context.Context, path string, mrs yara.MatchRules, c malconten
...
+		}
+	}
+
+	// Regardless of running an `analyze` or `scan`, adjust the overall risk if we deviated from overallRiskScore
+	// Scans will still need to drop medium or lower results
+	newRisk := highestBehaviorRisk(fr)
+	switch c.Scan {
+	case true:
+		switch {
+		case newRisk != overallRiskScore && newRisk >= 3:
```
tstromberg, line 575: Can this use a const of some sort? I don't remember if "3" is "HIGH" or "CRITICAL". I'm not sure why either needs to be hardcoded here to be honest, but I think it'd improve readability.

pkg/report/report.go
```diff
@@ -521,6 +517,72 @@ func Generate(ctx context.Context, path string, mrs yara.MatchRules, c malconten
...
+					break
+				}
+			}
+		}
+	}
+
+	// Regardless of running an `analyze` or `scan`, adjust the overall risk if we deviated from overallRiskScore
+	// Scans will still need to drop medium or lower results
+	newRisk := highestBehaviorRisk(fr)
+	switch c.Scan {
```
tstromberg, line 572: I don't think the nested switches are helping readability. Maybe something like:

```
if newRisk != overallRiskScore {
			overallRiskScore = newRisk
}

if c.Scan && newRisk < HIGH:
  return malcontent.FileReport{}, nil
}
```

Is this just checking if the command is "scan" vs "analyze"? If so, I don't think this function should need to know: all scan should need to do is set a default minRisk of "HIGH" and everything else should just work out.

egibs: Cleaned this up in [`660024a` (#487)](https://github.com/chainguard-dev/malcontent/pull/487/commits/660024ab189cc7f5e1bc854b07ddd0c7d50510ba).

pkg/report/report.go
egibs: Updated in [`660024a` (#487)](https://github.com/chainguard-dev/malcontent/pull/487/commits/660024ab189cc7f5e1bc854b07ddd0c7d50510ba).

pkg/report/report.go
egibs: Added two new fields for overrides in [`660024a` (#487)](https://github.com/chainguard-dev/malcontent/pull/487/commits/660024ab189cc7f5e1bc854b07ddd0c7d50510ba).

pkg/report/report.go
egibs: Moved to a new function in [`660024a` (#487)](https://github.com/chainguard-dev/malcontent/pull/487/commits/660024ab189cc7f5e1bc854b07ddd0c7d50510ba).

pkg/report/report.go
```diff
@@ -614,3 +630,41 @@ func highestMatchRisk(mrs yara.MatchRules) int {
...
+
+// handleOverrides modifies the behavior slice based on the contents of the override slice.
+func handleOverrides(fr malcontent.FileReport) []*malcontent.Behavior {
+	behaviorMap := make(map[string]*malcontent.Behavior, len(fr.Behaviors))
+	for _, b := range fr.Behaviors {
+		behaviorMap[b.RuleName] = b
+	}
+
+	for _, o := range fr.Overrides {
+		if b, exists := behaviorMap[o.Override]; exists {
```
egibs, line 658: I tested this out with a critical false positive and it works.

Override rule:
```
rule pulumi_shellcode : override {
  meta:
    malware_shellcode_hash = "low"
    description = "Ignore shellcode findings in the pulumi binary"
  strings:
    $pulumi = { 50 75 6C 75 6D 69 }
    $pulumi2 = { 70 75 6C 75 6D 69 }
  condition:
    any of ($pulumi*)
}
```

Result:
CleanShot 2024-10-04 at 13 16 39@2x

rules/net/dns-servers.yara
```diff
@@ -1,4 +1,4 @@
-rule go_dns_refs {
+rule go_dns_refs_local {
```
egibs, line 1: There were two `go_dns_refs` rules and there was a chance one would clobber the other:
- https://github.com/chainguard-dev/malcontent/blob/0acb2e0e98915c8fee1d6723dfbf1a29d861ed5c/rules/net/dns.yara#L1-L11
- https://github.com/chainguard-dev/malcontent/blob/0acb2e0e98915c8fee1d6723dfbf1a29d861ed5c/rules/net/dns-servers.yara#L1-L10

PR #490, opened 2024-10-06
Delay rule compilation and cache the results

tstromberg:
This PR moves rule compilation until the last possible moment and caches the result.

This ensures that path typos such as `mal scan ./nonexistent` return an error instantly instead of compiling the rules first.

Some example performance improvements:

* `mal xyz`: 40X faster
* `mal scan ./nonexistent`: 25X faster
* `make test`: 9X faster

This does not improve the result of `mal scan /path/to/malcontent-samples` however.


egibs (approved):
This is awesome. Funnily enough, I was thinking about caching compiled rules yesterday.

PR #491, opened 2024-10-06
Turn on prealloc linting rule, implement suggestions

egibs:
I was looking at linting rules to help with further optimizations of malcontent and noticed that we hadn't turned on `prealloc`.

Even for relatively small slices, this can have a positive impact: https://oilbeater.com/en/2024/03/04/golang-slice-performance/

This PR enables the rule and makes the appropriate modifications for both slices and maps (even though `prealloc` does not check the latter).

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -131,6 +131,7 @@ linters:
     - nilerr
     - noctx
     - nolintlint
+    - prealloc
     - predeclared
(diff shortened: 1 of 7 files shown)


tstromberg:
I'd eat my hat if this resulted in a noticeable performance increase for malcontent, but why not try :)

tstromberg (approved):

PR #509, opened 2024-10-09
Add jaraco py_dropper_chmod override

egibs:
Depends on: https://github.com/chainguard-dev/malcontent-samples/pull/12

This addresses one of the outstanding false positives we've seen.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,7 @@ fix: $(FIXERS)
 # END: lint-install ../malcontent
 
 SAMPLES_REPO ?= chainguard-dev/malcontent-samples
-SAMPLES_COMMIT ?= f56db86f52ef9b647943d9e3b6652a85ccb68266
+SAMPLES_COMMIT ?= 0ff28cbe99bc4610c58016faeb1a806a6e5cebbb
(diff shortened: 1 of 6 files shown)


egibs:
The test failure is expected until the `malcontent-samples` PR is merged and the commit is updated in this branch:
```
--- FAIL: TestSimple/python/clean/jaraco/__init__.py (0.00s)
        samples_test.go:145: scan failed: find: eval "python/clean/jaraco/__init__.py": lstat python/clean/jaraco/__init__.py: no such file or directory
```

tstromberg (changes requested):
Putting my "how can we make this more generally applicable" hat:

- I think `py_dropper_chmod` should be downgraded to HIGH; there is ample evidence that the rule is insufficiently tight enough to prevent real-world false-positives

- I think an override is still worthwhile to lower the priority of this alert to medium in known cases. I'd take the `not_*` in the `py_dropper_chmod` rules and make an override out of them.
- For this override, I think $comment2 is likely a sufficient hint to reduce the priority to medium. Alternatively, the name of the function may be less fragile: `func remove_readonly(`. It's unlikely to show up in real malware.

tstromberg (approved):

PR #510, opened 2024-10-10
rule tuning: make severities more appropriate

tstromberg:
risk tuning, mostly relating to "HIGH" findings in Wolfi. 

Add/fix testdata missing for clean Linux files.


rules/evasion/bitwise_math.yara
```diff
@@ -26,6 +26,7 @@ rule excessive_bitwise_math : high {
     $not_MD4 = "MD4"
     $not_algbase = "algbase" fullword
     $not_jslint = "jslint bitwise"
+	$not_includde = "#define "
```
egibs, line 29:
```suggestion
	$not_include = "#define "
```

rules/combo/stealer/crypto.yara
```diff
@@ -11,7 +11,9 @@ rule js_crypto_stealer : high {
     $url = /https{0,1}:\/\/[\w][\w\.\/\-_\?=\@]{8,64}/
 
 	$POST = "POST"
-
+	$not_webdav = "WebDAV certificate"
+	$not_letsencrypt = "Letsencrypt"
```
egibs, line 15: Would `/[Ll]etsencrypt/` work here?

egibs (approved):
Nice!

Would you mind looking at https://github.com/wolfi-dev/os/pull/30457?

The `gitaly` binary is 252MB and 66MB after being compressed which is still pretty large. 

Would it make sense to tune the `linux_multi_persist` rule to high or just add a quick override rule for that binary?

tstromberg:
> Would you mind looking at [wolfi-dev/os#30457](https://github.com/wolfi-dev/os/pull/30457)?

gitaly looks interesting.

> Would it make sense to tune the `linux_multi_persist` rule to high or just add a quick override rule for that binary?

For now, I think the rule should hit rare enough that we should do overrides; assuming the binary we are looking at isn't doing anything bad.

rules/combo/stealer/crypto.yara
tstromberg: It would, but it will would execute a tiny bit more slowly. If we end up with a 3rd option, then I'd combine it into a regexp.

egibs:
> gitaly looks interesting.

Indeed. AFAIK none of the previous releases triggered that rule.

Issue #500, opened 2024-10-07
diff: add --risk-change and --risk-increase flags


tstromberg:
When used as part of a CI/CD pipeline, you may not want to see a diff of every individual capability change. Sometimes you only want to know if the risk changes or increases.

Here is an initial description of the intended behavior:

* `--file-risk-change`: only show a diff if a file risk score changes (lower or higher). If a new file is added, show the capabilities of that file. Diff should include all capability changes, not just those that raised the file risk level.
* `--file-risk-increase`: same as the above, but only if the risk grows

These should stay compatible with `--min-file-risk`, so far instance, `--file-risk-increase --min-file-risk=critical` would only show diffs if a program becomes critical that was not previously critical.

PR #514, opened 2024-10-11
Add --file-risk-change and --file-risk-increase flags

egibs:
Closes: #500

This PR adds two new flags for diffs -- `--file-risk-change` and `--file-risk-increase`.

While the former is essentially what `diff` does right now, it has the distinct difference of filtering out diff reports for modified source and destination files with equal risk scores.

The latter will filter out diff reports for modified source and destination files where the source file risk score is equal to or greater than the destination file risk score.

The filtering occurs before any `d.Modified.Set` calls and after the `d.Added.Set` or `d.Removed.Set` calls in order to preserve unrelated file changes.

I also added four different test cases for each flag --
- Source -> Destination Increase
- Source -> Destination Decrease
- Source -> Destination No change
- Source -> Destination Decrease (`--file-risk-change`) or Increase (`--file-risk-increase`) for unrelated files


pkg/render/markdown.go
```diff
@@ -77,7 +77,9 @@ func (r Markdown) Full(ctx context.Context, rep *malcontent.Report) error {
 				mdRisk(modified.Value.RiskScore, modified.Value.RiskLevel))
 		}
 
-		fmt.Fprint(r.w, title+"\n\n")
+		if len(modified.Value.Behaviors) > 0 {
```
egibs, line 80: This would print `Changed: <path>` even if there were no behaviors to list.

test_data/refresh-testdata.sh
```diff
@@ -57,6 +57,62 @@ addq ${malcontent} --format=markdown \
...
+	macOS/2023.3CX/libffmpeg.dirty.dylib
+
+addq ${malcontent} --format=markdown \
+	-o "${test_data}/macOS/2023.3CX/libffmpeg.change_decrease.mdiff" \
+	--file-risk-change \
+	diff \
+	macOS/2023.3CX/libffmpeg.dirty.dylib \
+	macOS/2023.3CX/libffmpeg.dylib
+
+addq ${malcontent} --format=markdown \
```
egibs, line 74: This technically does nothing, but the filtering works:
```
time=2024-10-10T20:56:45.979-05:00 level=INFO source=.../malcontent/pkg/action/diff.go:323 msg="dropping result because diff scores were the same" paths="out/samples-0ff28cbe99bc4610c58016faeb1a806a6e5cebbb/macOS/2023.3CX/libffmpeg.dylib (2) out/samples-0ff28cbe99bc4610c58016faeb1a806a6e5cebbb/macOS/2023.3CX/libffmpeg.dylib (2)"
time=2024-10-10T20:56:45.979-05:00 level=INFO source=.../malcontent/pkg/action/diff.go:323 msg="dropping result because diff scores were the same" paths="out/samples-0ff28cbe99bc4610c58016faeb1a806a6e5cebbb/macOS/2023.3CX/libffmpeg.dylib (2) out/samples-0ff28cbe99bc4610c58016faeb1a806a6e5cebbb/macOS/2023.3CX/libffmpeg.dylib (2)"
```

egibs: It might make sense to do some diffs with different paths as well. I was testing that locally by copying the contents of `macOS` to a separate directory and then generating the diffs based on those paths.

tstromberg (approved):
Looks good. Can you also add these to the README?

PR #515, opened 2024-10-11
Add filesize condition to linux_multi_persist rule

egibs:
Follow-up for: https://github.com/wolfi-dev/os/pull/30457

This PR initially added a Gitaly override rule for `linux_persist_multi` but now adds a `filesize < 20MB` condition which will automatically ignore larger files (`gitaly` was ~270-280MB).


tstromberg (approved):
Looks good. I also wonder if the original rule should be guarded by a file size. IMHO, any Linux binary over 10-20MB is unlikely to be a persistence dropper.

egibs:
> Looks good. I also wonder if the original rule should be guarded by a file size. IMHO, any Linux binary over 10-20MB is unlikely to be a persistence dropper.

Good point. I'll rework the PR to add the filesize limit which will make the override moot.

rules/persist/linux_multi.yara
```diff
@@ -14,5 +14,5 @@ rule linux_multi_persist : critical {
     $bash_ref6 = "/etc/bashrc"
     $bash_ref7 = "/etc/bash"
   condition:
-    ($initd or $udev) and $crontab and any of ($bash*)
+    filesize < 20MB and ($initd or $udev) and $crontab and any of ($bash*)
```
egibs, line 17:
```
$ go run cmd/mal/mal.go analyze ./out/gitaly 
🔎 Scanning "./out/gitaly"
out/gitaly [🔥 HIGH]
```

Issue #512, opened 2024-10-10
Add risk values to simple output


tstromberg:
Simple output should include risk value at both the file and rule level. 

As it's mostly used in tests, we should optimize for easy diffing; here's an example:

```
# filename: critical
* rule/name: low
* rule/name: high
```

tstromberg:
https://github.com/chainguard-dev/malcontent/blob/1d85ca343bc452fbc3a396584b7cb9d0f766cbe7/pkg/render/simple.go#L31 is where the changes would go.

PR #516, opened 2024-10-12
Add risk levels to simple output

egibs:
Closes: #512

This PR adds a lowercase risk level string next to scanned paths and their behaviors.

I also added consistent formatting to make the output a little cleaner.


egibs:
Looking into why the tests are wanting to downgrade the severity of several test cases when refreshing the test data and manual scans return the correct risk level.

test_data/samples_test.go
```diff
@@ -135,12 +135,13 @@ func TestSimple(t *testing.T) {
...
-				Concurrency: runtime.NumCPU(),
-				IgnoreSelf:  false,
-				IgnoreTags:  []string{"harmless"},
-				Renderer:    simple,
-				RuleFS:      []fs.FS{rules.FS, thirdparty.FS},
-				ScanPaths:   []string{binPath},
+				Concurrency:           runtime.NumCPU(),
+				IgnoreSelf:            false,
+				IgnoreTags:            []string{"harmless"},
+				QuantityIncreasesRisk: true,
```
egibs, line 141: @tstromberg -- without this, some of our known malicious samples were registering as `high` instead of `critical`, but for some reason it's not happening outside of this branch which is interesting.

egibs: That said, we do use `QuantityIncreasesRisk: true` for the markdown tests but not the diff or JSON tests.

pkg/render/simple.go
```diff
@@ -40,7 +41,7 @@ func (r Simple) File(_ context.Context, fr *malcontent.FileReport) error {
 	})
 
 	for _, b := range bs {
-		fmt.Fprintf(r.w, "%s\n", b.ID)
+		fmt.Fprintf(r.w, "%s:%s%s\n", b.ID, strings.Repeat(" ", maxLength(bs)-len(b.ID)+1), strings.ToLower(b.RiskLevel))
```
tstromberg, line 44: Can you just make this a single space between the ID and the risk level? Otherwise, we'll end up with the same inscrutable diff problem that markdown samples have today.

tstromberg (changes requested):

pkg/render/simple.go
egibs: Updated in [`8395a2a` (#516)](https://github.com/chainguard-dev/malcontent/pull/516/commits/8395a2a8d5c6557d247e5db4de853a4aefd0989c).

tstromberg (approved):

Issue #327, opened 2024-07-05
Add detection for kSpreader artifacts


tstromberg:
https://blog.xlab.qianxin.com/8220-k4spreader-new-tool-en/

There are some particular shell invocations that seem of interest.

Issue #325, opened 2024-07-04
Add detection for kubo/injector


tstromberg:
https://github.com/kubo/injector

mentioned in https://cloud.google.com/blog/topics/threat-intelligence/uncovering-unc3886-espionage-operations/

PR #517, opened 2024-10-13
linux: alert tuning for k4spreader, injector, medusa, Sliver

### LINKED ISSUE #324: Add detection for Medusa preload
https://github.com/ldpreload/Medusa

Mentioned here: https://cloud.google.com/blog/topics/threat-intelligence/uncovering-unc3886-espionage-operations/

tstromberg:
fixes #324 #325 #327

unrelated change: I changed the chrome testdata file from .md to .simple to be easier to reason about.


rules/evasion/fake-process.yara
```diff
@@ -42,3 +43,17 @@ rule fake_systemd : critical {
...
+rule fake_process_names : high {
+	meta:
+		description = "mentions known fake process names"
+	strings:
+		$kdevchecker = "kdevchecker" fullword
+		$kworkerr = "kworkerr" fullword
+		$ksoftriqd = "ksoftriqd" fullword
+		$kdevtmpfsi = "kdevtmpfsi" fullword
+		$kthreaddk = "kthreaddk" fullword
+		$deamon = "deamon" fullword
```
egibs, line 56: Should this be `daemon` in both cases?

tstromberg: Nope - this is an intentional misspelling based on what I found in malware samples.

tstromberg:
@egibs - this could use another look. thanks!

test_data/javascript/clean/frequency_lists.js.simple
```diff
@@ -1,10 +1,12 @@
-# javascript/clean/frequency_lists.js: high
+# javascript/clean/frequency_lists.js: critical
```
egibs, line 1: Should we allow any clean samples to register as critical?

tstromberg: In general, I would say no - but until we can define filetype rules, I think it's OK to present the fact that the rule is overly broad.

What I would like to define for `combo/degrader/infection` is: `filetypes = "!js"`

egibs (approved):

PR #527, opened 2024-10-16
Add mlflow pypi_package_index override rule, allow for multiple rules per override

egibs:
The auto-generated PyPI package index in the `mlflow` package contains ~565,000 package names which match several of our [critical] rules.

This PR drops the criticals to low; however, I can go back and tweak the highs as well.

To facilitate this, I also added support for multiple rules per override so that multiple overrides aren't required for a single file.


test_data/linux/clean/pypi_package_index.json.simple
```diff
@@ -0,0 +1,196 @@
+# linux/clean/pypi_package_index.json: critical
```
egibs, line 1: This is only a critical because of `QuantityIncreasesRisk: true`.

tstromberg (approved):

Issue #506, opened 2024-10-08
Add platform-specific tag to rules


r0binak:
This is only necessary for use in certain environments.

tstromberg:
Spoke to @r0binak via e-mail, the gist here is to provide a way for folks who are only interested in rules that impact a single platform to filter by or ignore other platforms. For example, for a Linux CI/CD platform, you probably don't care about Windows hits.

What I propose is adding `GOOS`-style tags to our rules that only impact a subset of platforms:

- `linux`
- `darwin`
- `windows`
- `solaris`
- ...

For 3rd party rules, we'll need to apply some heuristics to auto-tag rules: for example, rules that contain `Win32` in their name should be tagged as Windows.

r0binak:
@tstromberg Please tell me when this is planned to be completed in full?

PR #530, opened 2024-10-18
Decrease false-positives across Ubuntu 24.04, add more OS-specific tagging

tstromberg:
Decreases false-positives found in /usr/lib across hundreds of files.

While waiting for tests to pass, I started adding tags for #506 as well.


rules/combo/exploit/breakout.yara
```diff
@@ -11,21 +13,23 @@ rule probable_container_breakout : high {
...
+    $x_notify_on_release = "notify_on_release"
+    $x_apparmor = "/sys/kernel/security/apparmor"
+    $x_mount_cgroup = "mount -t cgroup"
+    $x_rdma_cgroup = "rdma cgroup"
+    $x_cgroup_procs = "cgroup.procs"
+
+	$not_systemd = "SYSTEMD_PROC_CMDLINE"
   condition:
-    3 of them and not containerd
+    4 of them and not container_managers
```
egibs, line 32: Does this rule need a `none of ($not*)` condition to account for `$not_systemd`?

tstromberg: Good catch; done.

egibs (approved):

PR #534, opened 2024-10-21
Add override for filebeat misp_sample.ndjson.log

egibs:
This PR addresses a false positive we saw with the latest version of filebeat. 

The finding resides in a `.log` file that contains rows of arbitrary JSON content so I added several, somewhat unqiue field strings for the condition.

### DIFF
--- a/rules/false_positives/filebeat.yara
+++ b/rules/false_positives/filebeat.yara
@@ -1,0 +1,12 @@
+rule misp_mdjson : override {
+  meta:
+    description = "misp_sample.mdjson.log"
+    lvt = "medium"
+  strings:
+    $attribute = "Attribute"
+    $event = "Event"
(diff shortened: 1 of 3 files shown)


tstromberg (approved):

rules/false_positives/filebeat.yara
```diff
@@ -0,0 +1,12 @@
...
+  meta:
+    description = "misp_sample.mdjson.log"
+    lvt = "medium"
+  strings:
+    $attribute = "Attribute"
+    $event = "Event"
+    $galaxy = "Galaxy"
+    $shadow = "ShadowAttribute"
+  condition:
+    all of them
```
tstromberg, line 11: can we add some kind of filesize guardrails on this? It also lowers the scanning load for files that are outside of those limits.

PR #540, opened 2024-10-24
Add additional Wolfi false positve overrides

egibs:
This PR addresses the remaining critical false positives for Wolfi packages with the exception of `frp` and `sqlmap`.

I tried to keep the strings for each override generic enough and added `filesize` guards to each of the new overrides.


tstromberg (approved):
LGTM. I think there is a general theme here for follow-ups: `infection_killer` and `lvt` need to be made stricter.

PR #546, opened 2024-10-30
Format rule files with yara-x and add Workflow Check

egibs:
As of `yara-x` `v0.10.0`, there is now a `fmt` command which will keep all of our rule formatting consistent. 

This PR runs `yr fmt` on all of our first-party rules and also adds a quick Workflow Check to fail if the formatting command updates files and results in a dirty `git` state.


tstromberg (changes requested):
Awesome! Any chance you can integrate this into `make lint` and run that from the action?

Otherwise, we'll run into situations where the local lint tool and GHA tool disagree as the versions diverge.

Issue #555, opened 2024-10-31
GitHub Action to run `make update-third-party` nightly


tstromberg:
We should automate this so that our rules don't fall so far behind.

PR #557, opened 2024-10-31
Add Workflow to update third-party rules and PR the changes

egibs:
Closes: #555

This PR adds a Workflow that can be run ad-hoc or automatically on a weekly basis.

The Workflow will update the third-party rules and then refresh our test data to ensure that any changes are captured.


.github/workflows/third-party.yaml
```diff
@@ -0,0 +1,53 @@
+name: Update third-party rules
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 0 * * 1"
```
egibs, line 6: Totally open to changing this.

tstromberg: How about every 12 hours to start with? 

That way we don't get to where we accidentally issue a release with week-old rules.

egibs: Updated in [`f0e9393` (#557)](https://github.com/chainguard-dev/malcontent/pull/557/commits/f0e9393990af2a87c4dd066f06a0eb6f4399a690).

tstromberg (approved):

PR #558, opened 2024-11-01
MalwareBehaviorCatalog follow-up: less naming stutter, less slashes

tstromberg:
Slashes are now only in rule names to show directory hierarchy - this removes the - to / conversion.


tstromberg:
lint error is intentional and dumb: `artefacts` appears as a name in the YARAForge signatures.

egibs:
> lint error is intentional and dumb: `artefacts` appears as a name in the YARAForge signatures.

Can we start excluding certain words?

https://docs.getwoke.tech/ignore/#in-line-and-next-line-ignoring
⬇️ 
```go
func main() {
  fmt.Println("here is the whitelist") // wokeignore:rule=whitelist

  // wokeignore:rule=blacklist
  fmt.Println("and here is the blacklist")
}
```

tstromberg:
> > lint error is intentional and dumb: `artefacts` appears as a name in the YARAForge signatures.
> 
> Can we start excluding certain words?
> 
> https://docs.getwoke.tech/ignore/#in-line-and-next-line-ignoring ⬇️
> 
> ```go
> func main() {
>   fmt.Println("here is the whitelist") // wokeignore:rule=whitelist
> 
>   // wokeignore:rule=blacklist
>   fmt.Println("and here is the blacklist")
> }
> ```

Yeah, I think we should kill this linter and do it through golangci-lint instead though. The fact that we have lint checks that aren't run locally is dumb and frustrating.

egibs (approved):

PR #586, opened 2024-11-05
Fix 'none' severity findings breaking tests

### LINKED ISSUE #585: TestSimple/linux/clean/busybox: failing due to overridden "none" results
It looks like our tests are adding `none` results to the simple output for overridden rules, unlike `make refresh-sample-testdata`:

```
    --- FAIL: TestSimple/linux/clean/busybox (0.08s)
        samples_test.go:160: Simple output mismatch: (-want +got):
              (
              	"""
              	... // 63 identical lines
              	fs/watch: low
              	impact/ddos/raw_flooder: medium
            + 	impact/degrade/infection: none
            + 	impact/degrade/linux_paths: none
            + 	impact/remote_access/reverse_shell: none
              	lateral/scan/tool: medium
              	net/ip/addr: medium
              	... // 24 identical lines
              	"""
              )
```

Example: https://github.com/chainguard-dev/malcontent/actions/runs/11676935450/job/32514021648?pr=584#step:8:420

Blocks #584

egibs:
Closes #585

Our tests were not applying a consistent level of minimum risk. To avoid showing behaviors with `none` for their severity, this PR sets `MinFileRisk` and `MinRisk` to `1` for all tests. For reference, `TestJSON` was already using `MinRisk: 1`.


pkg/report/report.go
```diff
@@ -98,6 +98,8 @@ var dateRe = regexp.MustCompile(`[a-z]{3}\d{1,2}`)
 // Map to handle RiskLevel -> RiskScore conversions.
 var Levels = map[string]int{
 	"harmless":   0,
+	"ignore":     0,
```
tstromberg, line 101: I'd like to make sure that the API boundary is clear without any unanticipated surprises. The fact that the API returns dropped results would be a surprise to me.

What do you think about assigning the levels to be dropped by default (at least `ignore` and `none`) as `-1`? I'm iffy on `harmless`, it's probably OK as is.

tstromberg (changes requested):

pkg/report/report.go
egibs: Makes sense. I like that.

egibs: Updated in [`323a1e0` (#586)](https://github.com/chainguard-dev/malcontent/pull/586/commits/323a1e071a80e8508721031f29a89740fd03231e).

tstromberg (approved):

PR #593, opened 2024-11-06
re-organize samples + integration tests to improve caching

tstromberg:
This re-organizes tests for speed:

* re-use the cloned samples repo directory for new commits (just swap the commit out)
* re-use previously decompressed files
* remove the test_data file dance

To support this, this PR makes the following changes:

* The `out/samples-*` folder is now named `out/<name of org>/<name of repo>`
* The `samples/` folder is no longer used
* The `test_data` directory has been renamed `tests`
* `go test .` now works from the newly named `tests/` subdirectory
* `make test` is now split into `make integration` (slow integration tests) and `make test` (fast unit tests).


Makefile
```diff
@@ -75,41 +79,34 @@ fix: $(FIXERS)
...
+# unit tests only
 .PHONY: test
-test: prepare-samples
-	go test ./out/samples-$(SAMPLES_COMMIT)
+test:
 	go test ./pkg/...
 
+# integration tests only
+.PHONY: integration
+integration: out/$(SAMPLES_REPO)/.decompressed-$(SAMPLES_COMMIT)
```
egibs, line 107: Do we want to run these slower tests on merge to `main` or on a schedule?

tstromberg: oh, we definitely want to run integration tests for all PR's.Thanks for catching that!

egibs (approved):

PR #594, opened 2024-11-06
Add Kibana overrides

egibs:
This PR adds two override rules for as many Kibana files. The latter file is a minified JS file which triggers a lot of our rules which is built from [here](https://github.com/elastic/kibana/tree/main/x-pack/plugins/security_solution/scripts).


tstromberg (approved):

egibs:
I'll merge this after https://github.com/chainguard-dev/malcontent-samples/pull/27.

PR #601, opened 2024-11-07
scan: fix missing newline, make less noisy

tstromberg:
Fixes two UI glitches I ran into while scanning:

- If there was no evidence to report, we'd forget to write a newline :(
- Noisy well-intended error message about override rules. Added a TODO there instead.


pkg/report/report.go
```diff
@@ -464,7 +464,7 @@ func Generate(ctx context.Context, path string, mrs yara.MatchRules, c malconten
 				b.Override = append(b.Override, k)
 				fr.Overrides = append(fr.Overrides, b)
 			case !exists && override:
-				logger.Errorf("Override %s matched with no overridden rule\n", m.Rule)
+				// TODO: return error if override references an unknown rule name
```
egibs, line 467: I wasn't entirely sold on this output, so this works in the meantime 👍🏻

egibs (approved):
LGTM, looks like:
```go
func Generate(ctx context.Context, path string, mrs yara.MatchRules, c malcontent.Config, expath string, logger *clog.Logger)
```
needs to be
```go
func Generate(ctx context.Context, path string, mrs yara.MatchRules, c malcontent.Config, expath string, _ *clog.Logger)
```

Issue #489, opened 2024-10-06
analyze no longer streams output as results are generated


tstromberg:
When scanning a large directory, such as `/usr`, malcontent no longer outputs the results for files as they complete: it appears to wait until the end. This makes for poor UX as malcontent may go minutes with no output.

I suspect this regression came from the multi-threading changes.

The `scan` command seems to have the same limitation.

egibs:
Worked on this over the weekend and I'm close to a solution using priority queues.

PR #609, opened 2024-11-11
Add dynamic progress bar to improve scanning feedback

egibs:
I was thinking about how to address the feedback void between pressing `Return` and seeing the rendered file report(s). This void is especially long when scanning directories with large numbers of files.

Since rendering the file reports in a live manner such that they remain in a deterministic order is tricky due to concurrency, I instead landed on adding a progress bar which will update as a scan progresses.

This PR adds a simple, thread-safe progress bar that will update as each file or archive in a scan path is processed when running `mal analyze` or `mal scan`:
```
$ go run cmd/mal/mal.go analyze /usr/bin/
🔎 Scanning "/usr/bin/"
Processing discovered file(s)...  27% ▐▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒                    ▐ (441/1586) [6s:6s]
```

It's important to note that the bar will update across multiple lines if additional lines are written to `stdout`, so I moved several of our `logger.Errorf` statements in `scan.go` to `.Warnf` since we're returning the critical errors most of the time anyway. I also made this feature opt-out for use-cases where the progress bar will be rendered incorrectly (CI mostly).


pkg/action/scan.go
```diff
@@ -342,10 +344,19 @@ func recursiveScan(ctx context.Context, c malcontent.Config) (*malcontent.Report
 			}
 			return nil
 		}
+
+		if bar.State().Description != "" {
```
egibs, line 348: This was the cleanest way to check if a `bar` was configured. 

Using reflect did not work and checking if the State was "zero" by using `bar.State() == (progressbar.State{})` when passing in `bar := progressbar.ProgressBar{}` was showing a mismatch that would not evaluate to true. Either way, we'll always set the description when actually using a progress bar.

pkg/action/scan.go
```diff
@@ -482,3 +504,22 @@ func Scan(ctx context.Context, c malcontent.Config) (*malcontent.Report, error)
 	}
 	return r, nil
 }
+
+// newBar returns a configured progress bar for use in `analyze` and `scan` contexts.
+func newBar() progressbar.ProgressBar {
+	return *progressbar.NewOptions(
+		0,
+		progressbar.OptionClearOnFinish(),
```
egibs, line 512: This option ensures that the progress bar will disappear upon finishing to preserve the original output we expect.

tstromberg:
I like this conceptually, but we should ensure that the solution doesn't impact our ability to address https://github.com/chainguard-dev/malcontent/issues/489 - which is a root cause for the long delay for results. The API was designed to stream results as soon as they were available -  I think we lost that in the move to parallelism.

That said, I do think the progress bar is a great idea.

egibs:
This is at odds with the re-adding of streaming output in the terminal for `analyze`/`scan` in #617.

It's possible that we can use third-party UI library to store the scrolling output in one container and then the progress bar in another container but that seems like a lot of effort for a simple progress bar.

PR #617, opened 2024-11-12
Display scan results as soon as results are generated

egibs:
Closes: #489

Up until now (post-concurrency changes), we've been rendering results in a deterministic way by storing all of the file paths in a slice, sorting the slice, and then iterating through a map using the sorted keys. This prevents any output feedback while a scan is actually occurring and is especially apparent when scanning directories with many files.

This PR re-adds real-time streaming of results which will not be deterministic (which we probably shouldn't care about in general), but I added explicit sorting to two of our tests to ensure that the result data is what we expected. An unintended improvement of this PR is that `--err-first-hit` and `--err-first-miss` now work like they're supposed to.

I also verified that `--err-first-hit` and `--err-first-miss` also work with this implementation:
```
$ for i in (seq 1 3); go run cmd/mal/mal.go --err-first-hit analyze /usr/bin/; end
🔎 Scanning "/usr/bin/"
👋 "/usr/bin/SafeEjectGPU": matched requested condition
🔎 Scanning "/usr/bin/"
👋 "/usr/bin/SafeEjectGPU": matched requested condition
🔎 Scanning "/usr/bin/"
👋 "/usr/bin/SafeEjectGPU": matched requested conditione /Library/Application\ Support/BTServer/; end
$ for i in (seq 1 3); go run cmd/mal/mal.go --err-first-miss analyze /Library/Application\ Support/BTServer/; end
🔎 Scanning "/Library/Application Support/BTServer/"
👋 "/Library/Application Support/BTServer/pincode_defaults.db": matched requested condition
🔎 Scanning "/Library/Application Support/BTServer/"
👋 "/Library/Application Support/BTServer/pincode_defaults.db": matched requested condition
🔎 Scanning "/Library/Application Support/BTServer/"
👋 "/Library/Application Support/BTServer/pincode_defaults.db": matched requested condition
```

I also fixed up a data race condition, simplified how file reports are stored, and improved the output of `--stats` to make more sense. As part of the files map simplification, I added code to the JSON and YAML renderers to convert the `sync.Map` data to a format that works with the respective `Marshal` functions.


egibs:
Converting this to a draft to refactor the refactor to use standard Golang constructs. I also found a data race in `diff.go` which needs to be fixed.

pkg/action/archive_test.go
```diff
@@ -237,14 +238,21 @@ func TestScanArchive(t *testing.T) {
 	}
 
 	outBytes := out.Bytes()
+	sort.Slice(outBytes, func(i, j int) bool {
```
tstromberg, line 241: If I understand this: this is sorting the entire byte stream one character at a time.

If we're looking for help with non-determinism, I feel like we should convert the byte stream to a slice of strings and sort that instead. To be honest though, I think comparing the JSON output would be the correct answer for this test, rather than simple. 

Since this is an archive test, we really only need the list of files found in the JSON.

pkg/action/oci_test.go
```diff
@@ -44,14 +45,21 @@ func TestOCI(t *testing.T) {
...
 	td, err := os.ReadFile("testdata/scan_oci")
 	if err != nil {
 		t.Fatalf("testdata read failed: %v", err)
 	}
 	// Sort the loaded contents to ensure consistent ordering
+	sort.Slice(td, func(i, j int) bool {
+		return td[i] < td[j]
+	})
 	want := string(td)
+
```
tstromberg, line 62: Same comment here - I think simple is the wrong output format for validating that we can scan OCI files.

pkg/render/yaml.go
```diff
@@ -27,9 +27,32 @@ func (r YAML) File(_ context.Context, _ *malcontent.FileReport) error {
 }
 
 func (r YAML) Full(_ context.Context, rep *malcontent.Report) error {
-	// Drop the applied filters
-	rep.Filter = ""
-	yaml, err := yaml.Marshal(rep)
+	// Make the sync.Map YAML-friendly
+	type yamlReport struct {
```
tstromberg, line 31: This looks like a duplicate of `jsonReport` in `pkg/render/json.go` - use the same object?

tstromberg (approved):
Approved - though we should update oci_test and archive_test to use a more appropriate output format going forward.

pkg/action/archive_test.go
egibs: I was thinking the same thing. I'll make that change.

egibs: This will require some additional work in how we display the paths of files scanned within archives. Currently they have the entire temporary path which will never be the same across test runs so that needs to be removed when the reports are built.

egibs: Updated in [`f2d2599` (#617)](https://github.com/chainguard-dev/malcontent/pull/617/commits/f2d2599b9505783c903b658b6b396c1c9b19b786).

pkg/action/oci_test.go
egibs: Updated in [`f2d2599` (#617)](https://github.com/chainguard-dev/malcontent/pull/617/commits/f2d2599b9505783c903b658b6b396c1c9b19b786).

pkg/render/yaml.go
egibs: Updated in [`f2d2599` (#617)](https://github.com/chainguard-dev/malcontent/pull/617/commits/f2d2599b9505783c903b658b6b396c1c9b19b786).

PR #631, opened 2024-11-15
Improve detection of machO backdoors & stealers

tstromberg:
Based on studying some recent submissions to the Malware Bazaar.


egibs:
It's odd that `make integration` doesn't seem to be completing with the changes in this branch. Generating the test data should indicate that the rules work and produce results but the tests never finishing is weird.

tstromberg:
Yeah, it looks like there may be no fast way to match a long string of alphanumeric text. It turns out that compiled regexps are run across all files - not just the ones matching conditions:

https://github.com/Neo23x0/YARA-Performance-Guidelines

tstromberg:
commit incoming to improve performance ...

tstromberg:
updated rules are still slower than I like, but they are faster than they were.

there doesn't seem to be a fast way to detect a very large string with YARA :(

tstromberg:
finished in 6 minutes - not bad! I had to remove a query, but the best improvement was adding a nul character up front to find the start character.

egibs (approved):

Issue #638, opened 2024-11-17
`yara-x` rule compilation shows errors and warnings


egibs:
I was messing around with `yara-x`  today and tried to compile our rules. Some of our rules aren't compatible (`{` and `}` need to be escaped in regex patterns, for one, so I ran this with `--relaxed-re-syntax`), but a lot were displayed with warnings/suggestions:
```
error[E012]: duplicate rule `chromium_master_password`
 --> rules/credential/browser/chromium_master_password.yara:1:6
  |
1 | rule chromium_master_password: high {
  |      ^^^^^^^^^^^^^^^^^^^^^^^^ duplicate declaration of `chromium_master_password`
  |
 ::: rules/credential/browser/chromium_credit_cards.yara:1:6
  |
1 | rule chromium_master_password: critical {
  |      ------------------------ note: `chromium_master_password` declared here for the first time
  |
error[E012]: duplicate rule `sigaction_SIGALRM`
 --> rules/os/signal/handle-ALRM.yara:1:6
  |
1 | rule sigaction_SIGALRM: harmless {
  |      ^^^^^^^^^^^^^^^^^ duplicate declaration of `sigaction_SIGALRM`
  |
 ::: rules/os/signal/handle-WINCH.yara:1:6
  |
1 | rule sigaction_SIGALRM: harmless {
  |      ----------------- note: `sigaction_SIGALRM` declared here for the first time
  |
error[E012]: duplicate rule `listen`
 --> rules/net/socket/socket-listen.yara:1:6
  |
1 | rule listen: medium {
  |      ^^^^^^ duplicate declaration of `listen`
  |
 ::: rules/net/socket/pair.yara:1:6
  |
1 | rule listen: medium {
  |      ------ note: `listen` declared here for the first time
  |
error[E012]: duplicate rule `listen_tcp`
 --> rules/net/tcp/listen.yara:1:6
  |
1 | rule listen_tcp: medium {
  |      ^^^^^^^^^^ duplicate declaration of `listen_tcp`
  |
 ::: rules/net/tcp/connect.yara:1:6
  |
1 | rule listen_tcp: medium {
  |      ---------- note: `listen_tcp` declared here for the first time
  |
error[E012]: duplicate rule `elf_or_macho`
  --> rules/c2/addr/ip.yara:22:14
   |
22 | private rule elf_or_macho {
   |              ^^^^^^^^^^^^ duplicate declaration of `elf_or_macho`
   |
[...]

PR #640, opened 2024-11-17
Address yara-x compile findings

egibs:
Addresses: #638

Running `yr compile rules/` returned ~73 errors which were mostly duplicate rule names. Some of these rules were direct duplicates of each other; others were just needed a more appropriate (if longer) name.

In cases where `(elf or macho)` was used, the private rules needed to be renamed because the compiler thought we were referencing the import name (with a `struct` type).

Otherwise, I fixed a few strings and regenerated all of the test data to account for the updated rules.

To catch future rule warnings or errors, I added another CI check that will run `yr compile` on the entire rules directory which barely takes a second. If any findings exist, the Job will fail.


rules/impact/rootkit/rootkit.yara
```diff
@@ -44,34 +44,3 @@ rule linux_kernel_module_hide_self: critical linux {
   condition:
     filesize < 1MB and $register_kprobe and any of ($hide*)
 }
-
-rule funky_high_signal_killer: high {
```
egibs, line 48: This was a line-for-line duplicate.

rules/impact/rootkit/rootkit.yara
```diff
@@ -44,34 +44,3 @@ rule linux_kernel_module_hide_self: critical linux {
...
-
-  strings:
-    $odd_teen_sig = /kill -1[012346789]/ fullword
-    $high_sig     = /kill -[23456]\d/ fullword
-
-  condition:
-    filesize < 10MB and any of them
-}
-
-rule linux_rootkit_terms: critical linux {
```
egibs, line 60: This was a line-for-line duplicate.

rules/evasion/self_deletion/run_sleep_delete.yara
```diff
@@ -29,15 +29,3 @@ rule fetch_run_sleep_delete: critical {
   condition:
     filesize < 1KB and $url and $sleep and $rm and any of ($path*) and any of ($run*)
 }
-
-rule self_delete: high {
```
egibs, line 33: This was a line-for-line duplicate.

.github/workflows/style.yaml
```diff
@@ -117,22 +141,3 @@ jobs:
           version: v1.62.0
           args: --timeout=5m
 
-  lint:
```
egibs, line 120: This was doing nothing and we're already covering what it would handle via the other Jobs.

tstromberg (approved):

Issue #587, opened 2024-11-05
scan: too many regular expression fibers on acquia-toolbelt-2.4.1.gem


tstromberg:
When using `go run ./cmd/mal --err-first-miss scan $HOME/src/malware/Backstabbers-Knife-Collection/samples/rubygems/`:

```
=🔎 Scanning "/var/home/t/src/malware/Backstabbers-Knife-Collection/samples/rubygems/"

time=2024-11-05T14:41:10.354-05:00 level=ERROR source=/var/home/t/src/malcontent/pkg/action/scan.go:437 msg="scan error: scan: too many regular expression fibers" archivePath=/var/home/t/src/malware/Backstabbers-Knife-Collection/samples/rubygems/acquia-toolbelt/2.4.1/acquia-toolbelt-2.4.1.gem path=/tmp/acquia-toolbelt-2.4.1.gem2731929125/spec/cassettes/tasks/all_tasks.json
time=2024-11-05T14:41:10.355-05:00 level=ERROR source=/var/home/t/src/malcontent/pkg/action/scan.go:276 msg="unable to process /var/home/t/src/malware/Backstabbers-Knife-Collection/samples/rubygems/acquia-toolbelt/2.4.1/acquia-toolbelt-2.4.1.gem: report error: scan: too many regular expression fibers"
```

This seems to be returned by:

https://github.com/hillu/go-yara/blob/4f5cd55ec92e1cf17f11e2d74e232a178d1e2f58/error.go#L100C2-L100C27

Which in turn is likely caused by:

https://github.com/VirusTotal/yara/blob/548ba3d9c860e7b6c7286cbcf1f7e429cffe99f4/libyara/re.c#L1237

It's not a parallelization issue: setting `--jobs=1` 

I think this is just a YARA quirk maybe triggered by one of our rules? The file in question in 408K JSON file that is all on a single line.

tstromberg:
Based on research by @egibs - https://github.com/chainguard-dev/malcontent/pull/645 probably fixes this.

tstromberg:
confirmed that this is fixed at HEAD.

PR #643, opened 2024-11-18
Add flag to ignore problematic file extensions

egibs:
Relates to: #587

We're failing to scan certain files that are essentially one long string. Scanning these files with our compiled ruleset seems to be too much for the Yara library and results in this error:
```
💣 process: report error: scan: too many regular expression fibers
```

This PR adds a new flag to ignore certain file extensions that cause this error to work around this for the time being.

Example:
```
$ go run cmd/mal/mal.go --verbose --ignore-exts=.map analyze ~/Downloads/prism/usr/src/prism/node_modules/micri/dist/index.js.map
🔎 Scanning ".../Downloads/prism/usr/src/prism/node_modules/micri/dist/index.js.map"
time=2024-11-18T17:29:18.025-06:00 level=INFO source=.../repos/chainguard-dev/malcontent/pkg/action/scan.go:105 msg="skipping .../Downloads/prism/usr/src/prism/node_modules/micri/dist/index.js.map [.map]: ignored file extension" path.../Downloads/prism/usr/src/prism/node_modules/micri/dist/index.js.map
```

We'll continue looking into handling files like this in a more resilient way.

### DIFF
--- a/cmd/mal/mal.go
+++ b/cmd/mal/mal.go
@@ -52,6 +52,7 @@ var (
 	fileRiskChangeFlag        bool
 	fileRiskIncreaseFlag      bool
 	formatFlag                string
+	ignoreExtsFlag            string
 	ignoreSelfFlag            bool
 	ignoreTagsFlag            string
 	includeDataFilesFlag      bool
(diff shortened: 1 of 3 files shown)


egibs:
This might not be needed. I was able to locate the problematic string/rule.

PR #647, opened 2024-11-19
Use CachedRules in tests similarly to refresh

egibs:
We can leverage `CachedRules` across our parallel tests to save additional time like we do for `mal refresh`. It's nothing earth-shattering but it was saving ~5 seconds for me locally.

### DIFF
--- a/pkg/action/archive_test.go
+++ b/pkg/action/archive_test.go
@@ -233,7 +233,6 @@ func TestScanArchive(t *testing.T) {
 		MinFileRisk: 0,
 		MinRisk:     0,
 		Renderer:    r,
-		RuleFS:      rfs,
 		Rules:       yrs,
 		ScanPaths:   []string{"testdata/apko_nested.tar.gz"},
 	}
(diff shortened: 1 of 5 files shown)


pkg/action/archive_test.go
```diff
@@ -221,16 +221,23 @@ func TestScanArchive(t *testing.T) {
...
+	}
+
+	mc := malcontent.Config{
 		Concurrency: runtime.NumCPU(),
 		IgnoreSelf:  false,
 		MinFileRisk: 0,
 		MinRisk:     0,
 		Renderer:    r,
-		RuleFS:      []fs.FS{rules.FS, thirdparty.FS},
+		RuleFS:      rfs,
```
tstromberg, line 236: Why do we have to pass both Rules and RuleFS in?

tstromberg (changes requested):

pkg/action/archive_test.go
egibs: Updated that in [`1531cce` (#647)](https://github.com/chainguard-dev/malcontent/pull/647/commits/1531cce2dbfa64be7803ec382ee6925ecb17cecb).

tstromberg (approved):

Issue #648, opened 2024-11-19
"--stats scan" gives misleading statistics (analyze is fine)


tstromberg:
When you use `--stats` with `scan`, you can't get an idea of your hit rate as HIGH+CRIT always totals 100%. My guess is that it's because we drop misses before stat generation.

For example:

```
📊 Statistics
---
Files Scanned   10781 (0 skipped)
Total Risks     10781
---
⚠️  Risk Level Percentage
---
Risk Level    Percentage Count/Total
4/CRIT            89.46% 9645/10781
3/HIGH            10.54% 1136/10781
---
Number of behaviors      15758
```

egibs:
Yep, we return an empty file report in `report.Generate` when running a `scan`:
```go
if c.Scan && overallRiskScore < HIGH {
	return malcontent.FileReport{}, nil
}
```

I'll reason about a fix for this.

PR #649, opened 2024-11-19
Correctly calculate statistics when running scans

egibs:
Addresses #648

Running a `scan` would drop any file report whose risk was < `HIGH`. This threw off `--stats` since we'd only calculate the percentage of critical and high risks from the number of file reports matching those severities.

This PR tweaks the report and statistics logic so that we have the same number of files that `analyze` would have.

Example:
```
$ go run cmd/mal/mal.go --stats scan ./out/chainguard-dev/malcontent-samples/python/
...
📊 Statistics
---
Files Scanned   89 (48 skipped)
Total Risks     41
---
⚠️  Risk Level Percentage
---
Risk Level    Percentage Count/Total
4/CRIT            39.33% 35/89
3/HIGH             6.74% 6/89
---
Number of behaviors         72
...
```

`Files Scanned` matches what `analyze` sees:
```
$ go run cmd/mal/mal.go --stats analyze ./out/chainguard-dev/malcontent-samples/python/
📊 Statistics
---
Files Scanned   89 (0 skipped)
Total Risks     89
---
⚠️  Risk Level Percentage
---
Risk Level    Percentage Count/Total
2/MED             44.94% 40/89
4/CRIT            39.33% 35/89
1/LOW              8.99% 8/89
3/HIGH             6.74% 6/89
---
Number of behaviors        962
...
```

### DIFF
--- a/pkg/render/stats.go
+++ b/pkg/render/stats.go
@@ -19,22 +19,34 @@ func smLength(m *sync.Map) int {
 	return length
 }
 
-func riskStatistics(files *sync.Map) ([]malcontent.IntMetric, int, int) {
+func riskStatistics(c *malcontent.Config, files *sync.Map) ([]malcontent.IntMetric, int, int, int) {
 	length := smLength(files)
 
 	riskMap := make(map[int][]string, length)
 	riskStats := make(map[int]float64, length)
 
-	// as opposed to skipped files
 	processedFiles := 0
+	skippedFiles := 0
 	files.Range(func(key, value any) bool {
 		if key == nil || value == nil {
 			return true
 		}
 		processedFiles++
+
 		if fr, ok := value.(*malcontent.FileReport); ok {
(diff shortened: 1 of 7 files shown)


pkg/render/stats.go
```diff
@@ -52,16 +64,16 @@ func riskStatistics(files *sync.Map) ([]malcontent.IntMetric, int, int) {
...
 	sort.Slice(stats, func(i, j int) bool {
 		return stats[i].Value > stats[j].Value
 	})
 
-	return stats, total(), processedFiles
+	return stats, total(), processedFiles, skippedFiles
 }
 
-func pkgStatistics(files *sync.Map) ([]malcontent.StrMetric, int, int) {
+func pkgStatistics(_ *malcontent.Config, files *sync.Map) ([]malcontent.StrMetric, int, int) {
```
egibs, line 76: Left this here for future use.

PR #651, opened 2024-11-20
Add script to find missing testdata

tstromberg:
I use this when I add new samples. script emits the commands you need to add initial empty testdata for the refresh stage to fill.


tstromberg:
In the future, this should probably get a Makefile target that respects the Makefile variables for sample information - but figured I should at least get this up there.

egibs (approved):

PR #653, opened 2024-11-20
remove mantic mentions

tstromberg:
This will require moving our runners to Ubuntu from 22.04 to 24.04

### DIFF
--- a/.github/workflows/style.yaml
+++ b/.github/workflows/style.yaml
@@ -114,7 +114,7 @@ jobs:
 
   golangci-lint:
     name: golangci-lint
-    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


egibs (approved):

.github/workflows/style.yaml
```diff
@@ -132,11 +132,6 @@ jobs:
 
       - name: install libyara-dev
         run: |
-          sudo add-apt-repository -n -y "deb http://archive.ubuntu.com/ubuntu/ mantic main restricted universe multiverse"
```
egibs, line 135: We might have to change the runner for this Job to `ubuntu-24.04` until this is aliases to `ubuntu-latest`.

PR #665, opened 2024-11-26
Add new BubbleTea TUI renderer

egibs:
This PR adds a new, separate terminal renderer that leverages BubbleTea for a more interactive experience that can be accessed via `--format tui`. I also tweaked the normal terminal output to add additional color highlighting to strings.

This new renderer allows for scrolling/keyboard navigation within the viewport as well as searching for specific strings.

There are likely some small bugs in the implementation (e.g., boxes rendering incorrectly if the terminal window is too small), but for the most part it works as expected for `analyze`, `diff`, and `scan`. 

The main caveat when using this renderer is that [most] errors will be ignored to prevent breaking the interface.


tstromberg:
What do you think about naming this renderer "interactive"?

tstromberg (approved):

tstromberg:
(approved, other than the renderer name and new merge conflicts)

PR #673, opened 2024-12-03
Minor YARA rule tuning based on upcoming talk

tstromberg:
- Add queries to improve results for wiper/exfil examples presented
- Remove `collect/databases/sql` rule: too many falses-positives, including xz-utils.
- Remove `fullword` restriction from ssh private key detection as it doesn't work well with Go strings that aren't null terminated. 
- raise arbitrary_command_dev_null from medium to high

Fixes not shown in the presentation, but noticed while preparing for it:

- Lower criticality of looser "CRITICAL" rules to "HIGH".
- genericize `yarn` exception to `exec/cmd/npm_preinstall`


tstromberg:
adding @philroche @imjasonh so they can see how the sausage is made; admittedly this rule update ended up veering around a bit, but it's pretty typical of the type of rule tuning we do when we analyze a new set of malicious samples.

imjasonh (approved):

PR #694, opened 2024-12-10
Address recent, non-data file false positives

egibs:
This PR addresses many of the recent false positives for non-data files; however, there are likely a few more false positives to fix.

Two of the main improvements are the `probably_js` rule changes (which was alerting on files that were definitely not JavaScript) and turning off the `SECUINFRA_SUSP_Powershell_Base64_Decode` third-party rule which explicitly states that it can result in many false positives.


rules/anti-static/base64/function_names.yara
```diff
@@ -122,8 +122,9 @@ rule base64_php_functions_multiple: critical {
     $f_wp_nonce_field     = "wp_nonce_field" base64
     $f_wp_verify_nonce    = "wp_verify_nonce" base64
 
-    $not_comment     = "// processing instruction, e.g. <?php ?>"
-    $not_mongosh_php = { 3C 3F 70 68 70 00 00 00 01 0C 51 61 03 00 00 00 02 00 00 00 3F 3E }
```
egibs, line 126: This was a bit too specific and no longer applied to updated versions of `mongosh`.

rules/false_positives/trino.yara
```diff
@@ -0,0 +1,13 @@
...
+    description                                     = "launcher"
+    SIGNATURE_BASE_SUSP_ELF_LNX_UPX_Compressed_File = "high"
+
+  strings:
+    $go      = "go1.23"
+    $module  = "/Users/martin/go/pkg"
+    $partial = "martin/go"
+
+  condition:
+    filesize <= 1536KB and any of them
```
tstromberg, line 12: Should this be "all of them"? The go string is fairly broad.

Alternatively, I'm ok with changing the underlying  rule to "high", but maybe not until we switch wolfi to use differential analysis.

tstromberg (changes requested):

rules/false_positives/trino.yara
egibs: Each of the launchers only had one of the strings 😞

tstromberg: Ok. The post-compression strings are essentially going to be unpredictable. Just mark the underlying rule as "high" & remove the override. 

Critical is supposed to be reserved for cases where there is no question about maliciousness. Here we have a great counter-example. We can leave the critical findings for tampered UPX rules that this program did not hit.

egibs: Turned the third-party rule off in [`acc4401` (#694)](https://github.com/chainguard-dev/malcontent/pull/694/commits/acc44011543bc24207995781b5a74548f1e79d53).

We can just rely on our UPX rule which is already marked as high.

rules/anti-static/obfuscation/js.yara
```diff
@@ -3,9 +3,14 @@ import "math"
 private rule probably_js {
   strings:
     $f_function = /function\(\w{0,8}\)/
-    $f_const    = "const" fullword
-    $f_return   = "return" fullword
-    $f_var      = "var" fullword
+    $f_const    = /\bconst\b/
```
tstromberg, line 6: Isn't `"\b<text>\b"` and `"<text>" fullword` equivalent?

rules/anti-static/obfuscation/js.yara
```diff
@@ -3,9 +3,14 @@ import "math"
...
+    $f_return   = /\breturn\b/
+    $f_var      = /\bvar\b/
+    $f_let      = /\blet\b/
+    $f_if       = /\bif\b/
```
tstromberg, line 10: adding more 2-3 letter combinations is likely to raise our false-positive rate. Were there javascript files that you saw we weren't matching?

pkg/compile/compile.go
```diff
@@ -31,6 +31,9 @@ var badRules = map[string]bool{
 	"SIGNATURE_BASE_Reconcommands_In_File":                true,
 	"SIGNATURE_BASE_Apt_CN_Tetrisplugins_JS":              true,
 	"CAPE_Sparkrat":                                       true,
+	"SECUINFRA_SUSP_Powershell_Base64_Decode":             true,
+	// Rely on our first-party UPX rule
```
tstromberg, line 35: Either remove the comment or add it to the above base64 decoding one. Both have internal duplicates.

tstromberg (changes requested):

rules/anti-static/obfuscation/js.yara
egibs: `fullword` will match non-alphanumeric delimited strings, so things like `my-const` or `a-var` or `_return` will also match IIRC.

tests/linux/2021.FontOnLake/45E9.elf.simple
```diff
@@ -1,6 +1,6 @@
 # linux/2021.FontOnLake/45E9.elf: critical
 3P/elastic/rootkit_fontonlake: critical
-3P/sig_base/susp_elf_upx: critical
+anti-static/obfuscation/js: medium
```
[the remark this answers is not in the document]
egibs: Good point.

pkg/compile/compile.go
egibs: Updated in [`ec3ea08` (#694)](https://github.com/chainguard-dev/malcontent/pull/694/commits/ec3ea08e526c90802b024fb0bb349cc4c070dc7a).

rules/anti-static/obfuscation/js.yara
egibs: Good point -- tweaked in [`ec3ea08` (#694)](https://github.com/chainguard-dev/malcontent/pull/694/commits/ec3ea08e526c90802b024fb0bb349cc4c070dc7a).

egibs:
> > Two of the main improvements are the `probably_js` rule changes (which was alerting on files that were definitely not JavaScript)
> 
> I didn't notice any improvements here in the tests; any chance you have a sample you can add that validates that we've improved it?
> 
> Once #696 is fixed we can get rid of the `probably_js` rule.

Removing the extra additions and `fullword` removed a false positive. I was noticing non-JS files (not captured in the tests) matching so I wanted to tighten that up a bit.

rules/anti-static/obfuscation/js.yara
tstromberg: I did some testing, and as best as I can tell, `fullword` is equivalent to `\bword\b`. Tested with this rule:

```
rule const {
  strings:
    $fullword = "const" fullword
    $const_b_both = /\bconst\b/
    $const_s_both = /\sconst\s/
    $const_maybe_s_both = /\s{0,18}const\s/
    $const_anchor_space = /^\s{0,18}const\s/
  condition:
    any of them
}
```

```
# paste rule in
cat > /tmp/const.yara 

echo "const PI = 3.141569" >/tmp/const.js                          echo "/etc/const/" > /tmp/const.path
yara -s -w /tmp/const.yara /tmp/const.js
yara -s -w /tmp/const.yara /tmp/const.path
```

The strictest regex that works for this example is 
 `const_anchor_space`, followed by `const_maybe_s_both`.

When run against `/tmp/const.path`, the following match inadvertently:

```
0x5:$fullword: const
0x5:$const_b_both: const
```

rules/anti-static/obfuscation/js.yara
```diff
@@ -3,9 +3,9 @@ import "math"
 private rule probably_js {
   strings:
     $f_function = /function\(\w{0,8}\)/
-    $f_const    = "const" fullword
-    $f_return   = "return" fullword
-    $f_var      = "var" fullword
+    $f_const    = /\bconst\b/
```
tstromberg, line 6: looking at const/return/var, I think the strictest you can safely go is `\bword\s`; how about that?

rules/c2/tool_transfer/shell.yara
```diff
@@ -184,6 +186,10 @@ rule obsessive_dropper: critical {
     $cmd_echo      = "echo" fullword
     $cmd_chmod     = "chmod" fullword
 
+    $not_deckar01_bootstrap1 = "gem install bundler"
```
tstromberg, line 189: in addition, mark this rule `high`, as clearly there are false positives here.

rules/c2/tool_transfer/shell.yara
```diff
@@ -46,9 +46,11 @@ rule tool_chmod_relative_run_tiny: critical {
...
     $o_dot_slash = /\.\/[\$a-z]{1,2}[a-z\.\/\- ]{0,32}/ fullword
 
-    $not_copyright_comment = "# Copyright"
-    $not_source            = "source ./"
-    $not_apache_license    = "Apache License"
+    $not_copyright_comment  = "# Copyright"
+    $not_source             = "source ./"
+    $not_apache_license     = "Apache License"
+    $not_ruby_health_check1 = "chruby"
+    $not_ruby_health_check2 = "RUBY_VERSION"
```
tstromberg, line 53: in addition, mark this rule as `high`.

rules/data/embedded/embedded-base64-gzip.yara
```diff
@@ -28,12 +28,15 @@ rule base64_obfuscated_js: critical {
...
+    $f_substr          = "substr("
+    $f_join            = "join("
+    $f_function        = "function("
+    $not_events_js_map = "//# sourceMappingURL=events.js.map"
+    $not_jitsu1        = "admin-service-account@jitsu.com"
+    $not_jitsu2        = "jitsu-auth"
 
   condition:
-    filesize < 2MB and all of them and math.entropy(1, filesize) >= 5.0
+    filesize < 2MB and all of ($f*) and math.entropy(1, filesize) >= 5.0 and none of ($not*)
```
tstromberg, line 41: mark this rule as high

egibs: Updated in [`4260efd` (#694)](https://github.com/chainguard-dev/malcontent/pull/694/commits/4260efda23707a4e7d4a8539902b41a090f16684).

rules/c2/tool_transfer/shell.yara
egibs: Updated in [`4260efd` (#694)](https://github.com/chainguard-dev/malcontent/pull/694/commits/4260efda23707a4e7d4a8539902b41a090f16684).

rules/c2/tool_transfer/shell.yara
egibs: Updated in [`4260efd` (#694)](https://github.com/chainguard-dev/malcontent/pull/694/commits/4260efda23707a4e7d4a8539902b41a090f16684).

rules/anti-static/obfuscation/js.yara
egibs: Updated in [`4260efd` (#694)](https://github.com/chainguard-dev/malcontent/pull/694/commits/4260efda23707a4e7d4a8539902b41a090f16684).

tstromberg (approved):

PR #702, opened 2024-12-12
Fix non-tar bz2 extractions

egibs:
I noticed a few extraction errors for `.bz2` files:
CleanShot 2024-12-12 at 15 15 22@2x

These were the result of treating every `.bz2` file as a tar-based archive. This PR fixes the extraction by directly extracting non-tar.bz2 files:
```
$ go run cmd/mal/mal.go --include-data-files=true analyze ~/Downloads/test1.csv.bz2
🔎 Scanning "~/Downloads/test1.csv.bz2"
```

As opposed to:
```
🔎 Scanning "~/Downloads/test1.csv.bz2"
time=2024-12-12T15:05:24.317-06:00 level=ERROR source=.../repos/chainguard-dev/malcontent/pkg/action/scan.go:330 msg="unable to process ~/Downloads/test1.csv.bz2: extract to temp: failed to extract ~/Downloads/test1.csv.bz2: failed to read tar header: archive/tar: invalid tar header"
```


pkg/action/archive.go
```diff
@@ -140,6 +140,26 @@ func extractTar(ctx context.Context, d string, f string) error {
 		return nil
 	case strings.Contains(filename, ".bz2") || strings.Contains(filename, ".bzip2"):
```
tstromberg, line 143: While you are here, `.tbz` is a less common short extension form of `.tar.bz2`. If you add support for it, it might inspire a slightly different refactoring of this code.

pkg/action/archive.go
```diff
@@ -140,6 +140,26 @@ func extractTar(ctx context.Context, d string, f string) error {
...
 		br := bzip2.NewReader(tf)
+		// Extract non-tar archives directly
+		if !strings.HasSuffix(filename, ".tar.bz2") {
+			uncompressed := strings.TrimSuffix(filepath.Base(f), ".bz2")
+			uncompressed = strings.TrimSuffix(uncompressed, ".bzip2")
+			target := filepath.Join(d, uncompressed)
+			if err := os.MkdirAll(filepath.Dir(target), 0o755); err != nil {
```
tstromberg, line 150: Is world readable necessary? If not, maybe just `0700`?

tstromberg: instead of `filepath.Dir(target)`, there is already `d`; is it sufficient?

pkg/action/archive.go
```diff
@@ -140,6 +142,27 @@ func extractTar(ctx context.Context, d string, f string) error {
 		return nil
 	case strings.Contains(filename, ".bz2") || strings.Contains(filename, ".bzip2"):
 		br := bzip2.NewReader(tf)
+		// Extract non-tar archives directly
```
tstromberg, line 145: should code be located outside of the `extractTar` method then?

If you made this new code it's own method, it'd remove the complexity.

pkg/action/archive.go
```diff
@@ -140,6 +142,27 @@ func extractTar(ctx context.Context, d string, f string) error {
...
+		if !strings.HasSuffix(filename, ".tar.bz2") {
+			uncompressed := strings.TrimSuffix(filepath.Base(f), ".bz2")
+			uncompressed = strings.TrimSuffix(uncompressed, ".bzip2")
+			target := filepath.Join(d, uncompressed)
+			if err := os.MkdirAll(filepath.Dir(target), 0o755); err != nil {
+				return fmt.Errorf("failed to create directory for file: %w", err)
+			}
+
+			// #nosec G115
```
tstromberg, line 154: Can you add a comment as to what this directive is being used for?

pkg/action/archive.go
```diff
@@ -140,6 +142,27 @@ func extractTar(ctx context.Context, d string, f string) error {
...
+			out, err := os.OpenFile(target, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0o600)
```
tstromberg, line 155: I can't tell - but is this creating a new file in the same directory as the source file?  That may not work in all circumstances; for example, scanning a read-only mount or local directory where the user has read but not write access.

tstromberg (changes requested):

pkg/action/archive.go
egibs: It's landing in `/tmp` unless I'm misunderstanding:
```
time=2024-12-12T15:30:33.982-06:00 level=DEBUG source=.../repos/chainguard-dev/malcontent/pkg/action/archive.go:86 msg="extracting tar" dir=/var/folders/n6/xxn5d2zd3l1gghpx_7qppzs00000gn/T/test1.csv.bz21307193049
```

pkg/action/archive.go
egibs: Added in [`c32840a` (#702)](https://github.com/chainguard-dev/malcontent/pull/702/commits/c32840af4cba55983f37e6e233c37170267ba06b).

pkg/action/archive.go
egibs: Changed in [`c32840a` (#702)](https://github.com/chainguard-dev/malcontent/pull/702/commits/c32840af4cba55983f37e6e233c37170267ba06b).

pkg/action/archive.go
egibs: Updated in [`c32840a` (#702)](https://github.com/chainguard-dev/malcontent/pull/702/commits/c32840af4cba55983f37e6e233c37170267ba06b).

pkg/action/archive.go
egibs: Added in [`c32840a` (#702)](https://github.com/chainguard-dev/malcontent/pull/702/commits/c32840af4cba55983f37e6e233c37170267ba06b).

egibs:
Once this merges I'm going to have a follow up PR to clean up the `archive.go` organization since it's getting pretty lengthy now.

PR #708, opened 2024-12-16
Leverage yr scan --profile to tune slowest rules

egibs:
There were a few rules that `yr scan` showed as being slow in `main`, especially `decompress_base64_entropy`:
```
 1592 file(s) scanned in 7.4s. 559 file(s) matched.

«««««««««««« PROFILING INFORMATION »»»»»»»»»»»»

Slowest rules:

* rule                 : decompress_base64_entropy
  namespace            : default
  pattern matching     : 26.565195737s
  condition evaluation : 249.969µs
  TOTAL                : 26.565445706s

* rule                 : http_archive_url
  namespace            : default
  pattern matching     : 273.386026ms
  condition evaluation : 28.231µs
  TOTAL                : 273.414257ms

* rule                 : http_archive_url_higher
  namespace            : default
  pattern matching     : 273.386026ms
  condition evaluation : 21.056µs
  TOTAL                : 273.407082ms

* rule                 : osascript_window_closer
  namespace            : default
  pattern matching     : 271.315156ms
  condition evaluation : 18.015µs
  TOTAL                : 271.333171ms

* rule                 : osascript_quitter
  namespace            : default
  pattern matching     : 271.231338ms
  condition evaluation : 19.172µs
  TOTAL                : 271.25051ms
```

With these tweaks, the `decompress_base64_entropy` rule is now much, much faster with the other rules seeing a small improvement as well:
```
 1592 file(s) scanned in 4.0s. 559 file(s) matched.

«««««««««««« PROFILING INFORMATION »»»»»»»»»»»»

Slowest rules:

* rule                 : http_archive_url
  namespace            : default
  pattern matching     : 272.495941ms
  condition evaluation : 21.731µs
  TOTAL                : 272.517672ms

* rule                 : http_archive_url_higher
  namespace            : default
  pattern matching     : 272.495941ms
  condition evaluation : 14.334µs
  TOTAL                : 272.510275ms

* rule                 : decompress_base64_entropy
  namespace            : default
  pattern matching     : 122.998664ms
  condition evaluation : 7.835µs
  TOTAL                : 123.006499ms

* rule                 : osascript_window_closer
  namespace            : default
  pattern matching     : 122.534656ms
  condition evaluation : 14.036µs
  TOTAL                : 122.548692ms

* rule                 : osascript_quitter
  namespace            : default
  pattern matching     : 122.51377ms
  condition evaluation : 10.88µs
  TOTAL                : 122.52465ms
```

With these changes, `yr compile` runs cleanly and we get no performance warnings.


rules/anti-static/obfuscation/python.yara
```diff
@@ -449,10 +449,10 @@ rule decompress_base64_entropy: high {
     $f_decompress = "decompress("
     $f_eval       = "eval("
     $f_exec       = "exec("
-    $long_str     = /[\'\"][\+\w\/]{96}/
+    $long_str     = /[\'\"][\+\w\/]{,96}[\'\"]/
```
tstromberg, line 452: We don't have a test for this, but this rule will no longer work as expected: the intent of this rule is to find a base64-encoded string of at least 96 characters. I played with length ranges like {96,8192}, but it slowed the query down even more than just saying {96} and removing the quotes at the end.

rules/impact/degrade/app.yara
```diff
@@ -6,7 +6,7 @@ rule osascript_window_closer: medium {
     $c_osascript   = "osascript" fullword
     $c_tell        = "tell" fullword
     $c_application = "application" fullword
-    $c_app_name    = /\"\w[\.\w ]{0,24}\w\"/ fullword
+    $c_app_name    = /\"\w[\.\w]{0,24}\w\"/ fullword
```
tstromberg, line 9: `3,24` would be better targeting.

rules/anti-static/obfuscation/python.yara
egibs: Ah okay, that makes sense.

egibs: I still get a compiler warning with the changes in [`5ed3568` (#708)](https://github.com/chainguard-dev/malcontent/pull/708/commits/5ed3568d0e92415164899275755d13ea63785143) but it's much faster.

rules/anti-static/obfuscation/python.yara
```diff
@@ -440,19 +440,18 @@ rule decompress_base64_entropy: high {
...
-    filesize < 10MB and any of ($k*) and $b64decode and $long_str and any of ($f*)
+    $k_lzma         = "lzma"
+    $k_gzip         = "gzip"
+    $k_zlib         = "zlib"
+    $f_bytes        = "bytes("
+    $f_decode       = "decode("
+    $f_decompress   = "decompress("
+    $f_eval         = "eval("
+    $f_exec         = "exec("
+    $b64decode_long = /b64decode\(\"[\+\=\w\/]{96}/
```
tstromberg, line 451: While this modification will be faster, it's going to miss cases where they say:

```
y = "<long base64 encoded string>"
...
x = b64decode(y)
```

I'm OK with this change though until we have a testcase that necessitates the slower version.

tstromberg (approved):

rules/anti-static/obfuscation/python.yara
egibs: Ah, true. We may just want to eat the performance hit to cover all of the cases.

PR #716, opened 2024-12-17
build: reduce binary size by adding -s -w to ldflags

chenrui333:
Add `-s` (omit symbol table and debug info) and `-w` (omit DWARF debug info) to linker flags to minimize the size of the generated binaries.

```sh
## before

$ ls -alh /opt/homebrew/Cellar/malcontent/1.6.0/bin/mal
-r-xr-xr-x  1 rui  wheel    37M Dec 17 15:44 /opt/homebrew/Cellar/malcontent/1.6.0/bin/mal

## after

$ ls -alh /opt/homebrew/Cellar/malcontent/1.6.0/bin/mal
-r-xr-xr-x  1 rui  wheel    32M Dec 17 15:45 /opt/homebrew/Cellar/malcontent/1.6.0/bin/mal
```

relates to https://github.com/Homebrew/homebrew-core/pull/201530


egibs (approved):

egibs:
Thanks!

PR #723, opened 2024-12-18
Split up archive.go into type-specific files; add wider zlib support

egibs:
This PR splits up all of the archive-specific functions in `archive.go` into their own respective files for readability.

As part of this, I also added extension-agnostic `zlib` support depending on whether the file returns `&{.Z, application/zlib}` from `programkind.File`.

I also fixed a panic:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xd648e6]

goroutine 19400 [running]:
github.com/chainguard-dev/malcontent/pkg/action.extractGzip({0x2b5ce48?, 0xc0000ff6b0?}, {0xc001d680c0, 0x39}, {0xc0009b8240, 0x85})
	github.com/chainguard-dev/malcontent@v1.7.0/pkg/action/archive.go:249 +0x2c6
github.com/chainguard-dev/malcontent/pkg/action.extractArchiveToTempDir({0x2b5ce48, 0xc0000ff6b0}, {0xc0009b8240, 0x85})
	github.com/chainguard-dev/malcontent@v1.7.0/pkg/action/archive.go:653 +0x383
github.com/chainguard-dev/malcontent/pkg/action.processArchive({0x2b5ce48, 0xc0000ff6b0}, {0x8, 0x0, 0x0, 0x0, 0x0, 0x1, {0x0, 0x0, ...}, ...}, ...)
	github.com/chainguard-dev/malcontent@v1.7.0/pkg/action/scan.go:496 +0x134
github.com/chainguard-dev/malcontent/pkg/action.recursiveScan.func1({0xc0009b8240, 0x85})
	github.com/chainguard-dev/malcontent@v1.7.0/pkg/action/scan.go:328 +0xd0
github.com/chainguard-dev/malcontent/pkg/action.recursiveScan.func4()
	github.com/chainguard-dev/malcontent@v1.7.0/pkg/action/scan.go:436 +0xa2
golang.org/x/sync/errgroup.(*Group).Go.func1()
	golang.org/x/sync@v0.10.0/errgroup/errgroup.go:78 +0x50
created by golang.org/x/sync/errgroup.(*Group).Go in goroutine 19398
	golang.org/x/sync@v0.10.0/errgroup/errgroup.go:75 +0x96
```

This was happening because we were trying to print `ft.Ext` debug logs for files that returned `nil` from `programkind.File`.

### DIFF
--- a/pkg/action/archive.go
+++ b/pkg/action/archive.go
@@ -1,15 +1,8 @@
 package action
 
 import (
-	"archive/tar"
-	"archive/zip"
-	"compress/bzip2"
-	"compress/gzip"
-	"compress/zlib"
 	"context"
-	"errors"
 	"fmt"
-	"io"
 	"io/fs"
 	"os"
 	"path/filepath"
@@ -17,14 +10,8 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/cavaliergopher/cpio"
-	"github.com/cavaliergopher/rpm"
 	"github.com/chainguard-dev/clog"
 	"github.com/chainguard-dev/malcontent/pkg/programkind"
-
-	"github.com/egibs/go-debian/deb"
-
-	"github.com/ulikunitz/xz"
 )
 
 var archiveMap = map[string]bool{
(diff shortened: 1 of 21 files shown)


pkg/action/archive.go
```diff
@@ -85,512 +72,22 @@ func getExt(path string) string {
...
 func extractNestedArchive(
 	ctx context.Context,
 	d string,
 	f string,
 	extracted *sync.Map,
 ) error {
 	isArchive := false
-	ext := getExt(f)
-	if _, ok := archiveMap[ext]; ok {
+	// zlib-compressed files are also archives
```
egibs, line 82: These lines, L101-112, L152-162, and the additions in `zlib.go` are the only material changes in this PR. The rest are just moving existing code into new files (and updating the debug logging in `gzip.go`.

tstromberg:
It looks good; what do you think about moving this out of `action` and into an `archive` package? It feels like archive handling has grown up into its own feature at this point.

egibs:
> It looks good; what do you think about moving this out of `action` and into an `archive` package? It feels like archive handling has grown up into its own feature at this point.

I ran into some circular imports earlier but I can give it a try.

egibs:
> It looks good; what do you think about moving this out of `action` and into an `archive` package? It feels like archive handling has grown up into its own feature at this point.

Updated in [`43b922d` (#723)](https://github.com/chainguard-dev/malcontent/pull/723/commits/43b922de763b6b35e49f71c233fcd7978a6faf92).

I left `archive_test.go` in `action` since it uses `Scan` and `CachedRules`. I also moved the extension and archive map code to `programkind` since that seemed like a better fit.

tstromberg (approved):

PR #729, opened 2024-12-18
Avoid extracting joblib archives

egibs:
We're encountering failures when trying to extract files such as:
```
joblib/test/data/joblib_0.9.2_compressed_pickle_py35_np19.gz
```

While these appear to be valid gzip archives, they are actually just `application/octet-stream`/`data` files:
```
joblib/test/data/joblib_0.9.2_compressed_pickle_py35_np19.gz: data
```

This PR avoids extracting files like this if they contain common joblib extensions and have a MIME type of `application/octet-stream`.

To avoid creating temporary directories (and also scanning the equivalent of "" if `tmpRoot` is not created), I tweaked `processArchive` to return early if we detect these files. The nested extraction logic has also been updated to ignore joblib archives.


tstromberg:
I can't help but feel like hardcoding this exception introduces unnecessary complexity without a tangible benefit. Is this just to suppress output errors?

egibs:
> I can't help but feel like hardcoding this exception introduces unnecessary complexity without a tangible benefit. Is this just to suppress output errors?

Pretty much. Mainly since the files look like valid archives that should be extracting correctly.

egibs:
Closing this out for now since it's just to quiet down errors for unsupported archive types.

PR #733, opened 2024-12-20
Address more extraction edge cases; improve naming and consistency

egibs:
This PR addresses another subtle bug when extracting `.tar.gz` archives that are actually tar archives as well as other invalid `.gz` files and also adds better symlink handling.

I noticed this failure:
```
unable to process /tmp/malcontent-821299582/included-source/felix-ebpf-gpl.tar.gz: extract to temp: failed to extract /tmp/malcontent-821299582/included-source/felix-ebpf-gpl.tar.gz: failed to create gzip reader: gzip: invalid header
```

And inspected the file locally:
```
$ file felix-ebpf-gpl.tar.gz 
felix-ebpf-gpl.tar.gz: POSIX tar archive (GNU)
```

To fix this, I moved the `.gz` case statement below the `.tar.gz` and `.tgz` case statement so that it wouldn't preempt `ExtractTar` and also added a check to validate that a given file is actually a valid gzip file (and using the defaut tar reader if not).

With these changes, the file extracts and scans correctly:
```
🔎 Scanning "~/Downloads/calico-fips/included-source/felix-ebpf-gpl.tar.gz"
├─ 🟡 ~/Downloads/calico-fips/included-source/felix-ebpf-gpl.tar.gz ∴ /felix/bpf-gpl/bin/from_nat_info.o [MEDIUM]
│     ≡ command & control [MEDIUM]
│       🟡 addr/ip — mentions an IP and port:
│          client_ip, ctx_port, dst_port, get_port, host_ip, inner_ip, intf_ip, local_port, nat_ip, nat_port, orig_ip, orig_port, o…
│     ≡ credential [MEDIUM]
│       🟡 sniffer/bpf — BPF (Berkeley Packet Filter): bpf
│     ≡ discovery [MEDIUM]
│       🟡 multiple — collects system and network information: ip_addr, ipv46_addr, ipv4_addr, ipv6_addr
│     ≡ filesystem [MEDIUM]
│       🟡 path/relative — references and possibly executes relative path:
│          ./arp, ./bpf, ./conntrack_types, ./counters, ./failsafe, ./fib, ./include, ./jump, ./metadata, ./nat_lookup, ./nat_types…
│     ≡ networking [MEDIUM]
│       🟡 ip/addr — mentions an 'IP address': ip_addr
│       🟡 ip/host_port — connects to an arbitrary host:port: host(revnat->port, host_to_ctx_port
│       🔵 ip/icmp — ICMP (Internet Control Message Protocol), aka ping
│       🟡 socket/raw — send raw and/or malformed IP packets: IPPROTO_RAW
│       🔵 socket/send — send a message to a socket
│     ≡ persistence [LOW]
│       🔵 kernel_module/symbol_lookup — bpf: BPF
│
├─ 🟡 ~/Downloads/calico-fips/included-source/felix-ebpf-gpl.tar.gz ∴ /felix/bpf-gpl/bin/to_lo_fib_info_co-re.o [MEDIUM]
│     ≡ command & control [MEDIUM]
│       🟡 addr/ip — mentions an IP and port:
│          client_ip, ctx_port, dst_port, func_ip, get_port, host_ip, inner_ip, intf_ip, local_port, nat_ip, nat_port, orig_ip, ori…
│     ≡ credential [MEDIUM]
│       🟡 sniffer/bpf — BPF (Berkeley Packet Filter): bpf
│     ≡ discovery [MEDIUM]
│       🟡 multiple — collects system and network information: ip_addr, ipv46_addr, ipv4_addr, ipv6_addr
│     ≡ filesystem [MEDIUM]
│       🟡 path/relative — references and possibly executes relative path:
│          ./arp, ./bpf, ./conntrack_types, ./counters, ./failsafe, ./fib, ./ifstate, ./include, ./jump, ./nat_lookup, ./nat_types,…
│     ≡ networking [MEDIUM]
│       🟡 ip/addr — mentions an 'IP address': ip_addr
│       🟡 ip/host_port — connects to an arbitrary host:port: host source port, host(revnat->port, host_to_ctx_port
│       🔵 ip/icmp — ICMP (Internet Control Message Protocol), aka ping
│       🟡 ip/syncookie — references SYN cookies, used to resist DoS attacks: syncookie
│       🟡 socket/raw — send raw and/or malformed IP packets: IPPROTO_RAW
│       🔵 socket/send — send a message to a socket
│     ≡ persistence [LOW]
│       🔵 kernel_module/symbol_lookup — bpf: BPF
│
├─ 🔵 ~/Downloads/calico-fips/included-source/felix-ebpf-gpl.tar.gz ∴ /felix/bpf-gpl/include/libbpf/include/uapi/linux/if_xdp.h [LOW]
│     ≡ anti-static [LOW]
│       🔵 obfuscation/bitwise — uses bitwise math: 1 << 0, 1 << 1, 1 << 2, 1 << 3
│     ≡ networking [LOW]
│       🔵 socket/send — send a message to a socket: sendto
│
...
```


egibs:
Uncovered a few more edge cases while working on this, namely:
- `.7.gz` files packaged in Debian that are actually just ASCII files
- symlink handling within tar archives wasn't working as expected with nonexistent files

pkg/archive/archive.go
```diff
@@ -193,10 +193,10 @@ func ExtractionMethod(ext string) func(context.Context, string, string) error {
 	switch ext {
 	case ".jar", ".zip", ".whl":
 		return ExtractZip
-	case ".gz":
-		return ExtractGzip
 	case ".apk", ".gem", ".tar", ".tar.bz2", ".tar.gz", ".tgz", ".tar.xz", ".tbz", ".xz":
```
stevebeattie, line 196: Would it make sense to leave a comment here to indicate that added compression cases should go below this one?

egibs: Probably. The ordering can definitely cause issues if one extension is a substring of another.

egibs: Added in [`a31d5f6` (#733)](https://github.com/chainguard-dev/malcontent/pull/733/commits/a31d5f6aaa4dfbe494a6eccc016e0dc9e6751083).

stevebeattie:
> I noticed this failure:
> 
> ```
> unable to process /tmp/malcontent-821299582/included-source/felix-ebpf-gpl.tar.gz: extract to temp: failed to extract /tmp/malcontent-821299582/included-source/felix-ebpf-gpl.tar.gz: failed to create gzip reader: gzip: invalid header
> ```
> 
> And inspected the file locally:
> 
> ```
> $ file felix-ebpf-gpl.tar.gz 
> felix-ebpf-gpl.tar.gz: POSIX tar archive (GNU)
> ```

FYI, this specific instance looks to be a bug in the wolfi package for calico, in the build here https://github.com/wolfi-dev/os/blob/4e9ab93174c17cae8e9defb012b4a731621be1af/calico-3.29.yaml#L206-L212 ; it's trying to replicate the gpl ebpf tarball build from the upstream build process at https://github.com/projectcalico/calico/blob/810091a6acf98ecc891c5bc3c07374a786d68743/node/Makefile#L286-L290 but it dropped the `xz` compression without adding gzip compression, despite the target rename.

> - .7.gz files packaged in Debian that are actually just ASCII files

Do you have specific examples? The only example I could find on a running Ubuntu system was `/var/lib/dpkg/alternatives/builtins.7.gz` which is where the current configuration info is stored for `/etc/alternatives/builtins.7.gz` (I can understand the intent of this alternatives link, but it doesn't seem to be used by anything but the bash packages).

Overall, though probably outside the scope of this PR, I'd like to see such files be flagged by malcontent, as they are pretty rarely legit, they probably are either:

- bugs in the upstream or the downstream packaging (e.g. calico),
- the extremely rare legit case (alternatives of compressed files, test cases exercising these kinds of issues),
- or something trying to look like something it's not, which is a bit suspicious.

Thanks.

egibs:
> Do you have specific examples?

The example you found was the one I came across. TBD if there are others in other distros. 

I like the idea of flagging files for further review but the changes in this PR will at least avoid trying to extract archives that aren't valid.

pkg/archive/deb.go
```diff
@@ -74,21 +74,21 @@ func ExtractDeb(ctx context.Context, d, f string) error {
...
 			}
-			// Ensure that symlinks are not relative path traversals
-			// #nosec G305 // L208 handles the check
-			linkReal, err := filepath.EvalSymlinks(filepath.Join(d, header.Linkname))
+			linkReal, err := filepath.EvalSymlinks(fullLink)
 			if err != nil {
 				return fmt.Errorf("failed to evaluate symlink: %w", err)
 			}
-			if !IsValidPath(linkReal, d) {
+			if !IsValidPath(target, d) {
```
stevebeattie, line 88: I don't think this change is correct; I believe the intent here is to check that the destination of the symlink (what it points to) is within the temporary extraction directory.

However, we should also probably verifying that `isValid` is applied on the files, it seems that that's missing for the deb case:
https://github.com/chainguard-dev/malcontent/blob/3b499254a5714e4cd1cb95f05f110909cb70b849/pkg/archive/deb.go#L48-L50

unlike the tar case:

https://github.com/chainguard-dev/malcontent/blob/3b499254a5714e4cd1cb95f05f110909cb70b849/pkg/archive/tar.go#L103-L108

pkg/archive/tar.go
```diff
@@ -133,21 +144,21 @@ func ExtractTar(ctx context.Context, d string, f string) error {
...
 			}
-			// Ensure that symlinks are not relative path traversals
-			// #nosec G305 // L208 handles the check
-			linkReal, err := filepath.EvalSymlinks(filepath.Join(d, header.Linkname))
+			linkReal, err := filepath.EvalSymlinks(fullLink)
 			if err != nil {
 				return fmt.Errorf("failed to evaluate symlink: %w", err)
 			}
 			if !IsValidPath(target, d) {
 				return fmt.Errorf("symlink points outside temporary directory: %s", linkReal)
```
stevebeattie, line 159: So before this proposed change, the earlier os.Readlink() call would fail (because the symlink hadn't been created yet), and so for symlinks, they would never get to this. However, this check here should be changed to `if !IsValidPath(linkReal, d)` to ensure that the destination of the symlink is within the extraction directory. Without it, a symlink that points outside the directory will be created:

```
$ git checkout -q egibs/fix-tar-gz-edge-case
$ git show --no-patch --oneline HEAD
581cbe90 (HEAD, egibs/fix-tar-gz-edge-case) Add .gzip for completeness, move above .zip
$ go build  ./cmd/mal/
$ tar tvf ~/tmp/symlink.tar.gz
drwxrwxr-x user/user 0 2024-12-31 13:14 deleteme2/
lrwxrwxrwx user/user 0 2024-12-31 13:14 deleteme2/ls -> ../../../../../../../usr/bin/ls
$ ./mal --format=simple analyze ~/tmp/symlink.tar.gz
# /home/user/tmp/symlink.tar.gz ∴ /usr/bin/ls: medium
c2/addr/url: low
c2/tool_transfer/arch: low
data/compression/lzma: low
discover/system/hostname: low
evasion/file/prefix: medium
exec/shell/TERM: low
fs/link_read: low 
net/url/embedded: low
process/name_set: medium
```

stevebeattie (changes requested):

pkg/archive/tar.go
egibs: Oh, yea, good catch. It's hard to keep `target`,`linkReal, and `header.Linkname` straight.

pkg/archive/deb.go
egibs: Good catch -- this was definitely applied inconsistently. `zip.go` used it but I'll add it to the other methods as well.

pkg/archive/tar.go
stevebeattie: Given that the checks for symlinks are the same for both debs and tarballs, it seems likely that the validation could probably be abstracted into a helper function, so that there's not duplicated code that needs to be kept consistent.

egibs: Added in [`c0d9a99` (#733)](https://github.com/chainguard-dev/malcontent/pull/733/commits/c0d9a99b2e8bdbbe5cc0106891dbc1b355663f97).

I simplified the permissions; directories are `0o700` for traversal, everything else is `0o600`.

Additionally, `handleFile` will error if we exceed `maxBytes` (this was not evident prior) and `handleSymlink` will ensure that existing symlinks are cleaned up before we create them.

stevebeattie (approved):
Thanks for making these changes, this looks much better.

Issue #497, opened 2024-10-07
action: refactor recursiveScan


tstromberg:
This function has grown quite complex:

```
// recursiveScan recursively YARA scans the configured paths - handling archives and OCI images.
//
//nolint:gocognit // ignoring complexity of 101 > 98
func recursiveScan(ctx context.Context, c malcontent.Config) (*malcontent.Report, error) {
	logger := clog.FromContext(ctx)
	logger.Debug("recursive scan", slog.Any("config", c))
	r := &malcontent.Report{
		Files: orderedmap.New[string, *malcontent.FileReport](),
	}
	if len(c.IgnoreTags) > 0 {

```

It's at 169 lines of nested code. I'm guessing we can break something out to get it under 100 lines again.

Issue #227, opened 2024-05-17
Port malcontent to YARA-X


tstromberg:
YARA now has official Go bindings:

https://virustotal.github.io/yara-x/blog/yara-is-dead-long-live-yara-x/

There are only 2-3 calls into the existing library we use - https://github.com/hillu/go-yara - but I'm sure the data structure it returns has some differences that will need to be resolved.

Here's where I would start making the change:

https://github.com/chainguard-dev/bincapz/blob/996ba4c2c19d42694937f9b50af3bbc74b2ca3af/pkg/compile/compile.go#L59

This function, which converts YARA results into an internal FileReport struct will also need changes:

https://github.com/chainguard-dev/bincapz/blob/996ba4c2c19d42694937f9b50af3bbc74b2ca3af/pkg/report/report.go#L275

egibs:
Looks like [this](https://virustotal.github.io/yara-x/docs/api/c/c-/#building-the-c-library) will be a pre-req for working with Yara-X.

I ran into issues running `cargo install cargo-c` on macOS and had to run:
```
cargo clean && PATH="/usr/bin:$PATH" cargo install cargo-c
```
to avoid using `split` from Brew's `binutils`:
```
❯ which strip
/opt/homebrew/opt/binutils/bin/strip
```

tstromberg:
Ouch. With this change, it sounds like we'd be taking on two large dependencies that we didn't have before. I think that we should save this change for a little while - at least until YARA-X is available in homebrew and the mainstream Linux distros.

Closing this for now - but I expect we'll likely resurrect it by the end of the year.

egibs:
Found an example of Yara-X usage in Go: https://github.com/zz1gg/threatchecker/blob/main/utils/yara_x.go and used it as a reference for a very WIP feature branch.

There's a way to ignore modules when compiling rules but it doesn't seem like we get much control over warnings or errors like we do with the current package.

Aside from that, it looks like most everything else should be accessible except for Tags which don't seem to be a concept with Yara-X (at least as of version `0.5.0`).

fproulx-boostsecurity:
@tstromberg it's in homebrew now (https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/y/yara-x.rb)

tstromberg:
That's great news! @egibs has also been working on upstream patches that should soon make it possible to port malcontent to YARA-X.

PR #734, opened 2024-12-22
Migrate from go-yara to yara-x; improve performance and readability

egibs:
Closes: #227
Closes: #497

After hacking around with yara-x locally last night, the performance gains over Yara are definitely noticeable (usually 2x faster at minimum). Previous versions seemed to be slower (at least `0.10.0`) but my earlier testing may have been bugged/flawed. That said, I wanted to get this rather large refactor up to test the CI experience again since we have to build the API from scratch, but when refreshing the sample data locally I was down to about ~24-25 seconds with the integration tests taking about 27 seconds (on my M1 Pro MBP).

The main limitation with `yara-x` is that it does not expose any functionality to turn off problematic rules. To handle this, I added functionality to remove rules prior to compilation so that they are not evaluated by the scanner.

To help with CPU overhead, the PR also adds a pool of scanners that can be re-used across files. Previously, we always GC'd the active scanner after each scan which had a sizable impact, especially for scan paths containing many files.

This PR also cleans up `recursiveScan` and fixes the behavior of `longestUnique` and splits out path-related functions into a new `path.go` file. Additionally, `processArchive` will now concurrently scan extracted files which was previous serial. With this change, I can scan the OpenJDK package which extracts roughly 136,000 files in ~70-90 seconds.

Outside of the `longestUnique` changes, the final, rendered output is essentially 1:1 with the current implementation.

Edit: this PR is about 3-4x faster in GHA with 8-core runners (even when running in a container):
```
go test -timeout 0 ./tests/...
ok  	github.com/chainguard-dev/malcontent/tests	49.979s
```
> This is usually around ~170s.

The tests and `golangci-lint` jobs now run in a Wolfi container to avoid compiling the yara-x C API each time the Workflows run; this more than halves the runtime of each job (5+ minutes down to ~2 minutes).


.github/workflows/go-tests.yaml
```diff
@@ -17,7 +17,21 @@ permissions:
 jobs:
   test:
     runs-on: mal-ubuntu-latest-8-core
-    container: cgr.dev/chainguard/wolfi-base:latest
+    container:
+      image: cgr.dev/chainguard/wolfi-base@sha256:eeb70e74e2ac07d3c80a30150bf473970c8b51a57f06daef3e4d065ac52489bc
```
egibs, line 21: I wanted to wait until CI was running consistently before locking everything down.

README.md
```diff
@@ -118,16 +118,19 @@ The analyze mode emits a list of capabilities often seen in malware, categorized
 Requirements:
 
 * [go](https://go.dev/) - the programming language
-* [yara](https://virustotal.github.io/yara/) - the rule language
+* [yara-x](https://virustotal.github.io/yara-x/) - the rule language
 * [pkgconf](http://pkgconf.org/) - required by Go to find C dependencies, included in many UNIX distributions
 
-Linux or macOS users can run the following command to install the necessary dependencies, other than Go:
+`yara-x` requires an underlying C API to function. For Wolfi users, this can be installed by running `apk add yara-x-compat`.
```
tstromberg, line 124: This aside feels unhelpful as few if any users will install this onto Wolfi from source code. Can we provide a macOS example instead?

README.md
```diff
@@ -118,16 +118,19 @@ The analyze mode emits a list of capabilities often seen in malware, categorized
...
- || sudo dnf install yara-devel || sudo pacman -S yara \
- || sudo zypper install yara
+For other devices, reference the documentation here for installation instructions: https://virustotal.github.io/yara-x/docs/api/c/c-/#building-the-c-library
+
+Running `cargo cinstall -p yara-x-capi --release` may encounter permission denied errors. If this is the case, run the following:
 ```
+cargo cinstall -p yara-x-capi --release --prefix=/tmp/yara-x
+```
+
+And then copy the resulting files from `/tmp/yara-x` to `/usr/local/lib` and `/usr/local/include` (or the preferred `PKG_CONFIG_PATH`/`LD_LIBRARY_PATH` location(s)).
```
tstromberg, line 133: This feels kind of gross. Is there a separate build/install step in cargo? It'd be nice if we could tell folks to use `sudo cinstall ... --prefix=/usr/local`

tstromberg (approved):
Approved for code review, but the install instructions could use some more love.

README.md
egibs: I can test that out. IIRC `sudo` failed because it was looking for a root installation of `cargo`/`cinstall`.

egibs: Updated in [`0270283` (#734)](https://github.com/chainguard-dev/malcontent/pull/734/commits/0270283ae78bf130d4f2d6f1920601af62c641cc).

I tested the `sudo` command and it worked without any non-intuitive copying.

PR #739, opened 2024-12-29
 refactor recursiveScan function

Ritish134:
## Breakdown of recursiveScan function: #497 

- Created `initializeReport` function to initialize and configure the `malcontent.Report` object.
- I think Instead of relying on `sync.Once` to ensure that a match (or miss) is processed only once, I Created `setMatchHandling` function to do that, basically in this : 
Only the first goroutine to send a result to this channel succeeds due to the buffered capacity of 1.
```matchChan := make(chan matchResult, 1)```

`context.WithCancel` is used to signal all concurrent goroutines to stop further processing once a match is found.

Let me know your thoughts on this @tstromberg 

- the logic for handling OCI image extraction (including the `imageURI` and `ociExtractPath` variables) moved into the helper function `prepareScanPath`.

- The call to `findFilesRecursively` is now encapsulated within `processPaths`. This function is responsible for finding files and managing their subsequent processing.

- Concurrency Handling logic moved to `processWithConcurrency`

- File-specific processing logic into a separate function (`processSingleFile`)

- `storeFileReport` to store individual file reports and
`storeArchiveReports` For storing multiple file reports from a sync.Map

### DIFF
--- a/pkg/action/scan.go
+++ b/pkg/action/scan.go
@@ -264,227 +264,226 @@ func CachedRules(ctx context.Context, fss []fs.FS) (*yara.Rules, error) {
 	return compiledRuleCache, err
 }
 
+type matchResult struct {
+	fr  *malcontent.FileReport
+	err error
+}
 // recursiveScan recursively YARA scans the configured paths - handling archives and OCI images.
 //
 //nolint:gocognit,cyclop,nestif // ignoring complexity of 101,38
 func recursiveScan(ctx context.Context, c malcontent.Config) (*malcontent.Report, error) {
 	logger := clog.FromContext(ctx)
-	r := &malcontent.Report{
-		Files: sync.Map{},
-	}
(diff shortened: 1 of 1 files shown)


egibs:
Hey @Ritish134 -- appreciate the effort here! It seems like we were operating on the same wavelength with the proposed changes; however, I actually included a version of this in #734 as part of the larger migration to a newer Yara API and didn't see this until now.

As far as refactoring goes, there are some other areas that could use this attention, namely `Generate` in `report.go` and improving the performance/organization of `File` in `programkind.go`. 

Let me know if you're interested in either of these. I'm happy to log Issues with further details!

Ritish134:
Thanks! @egibs for the update. Sure, I would definitely be interested in working on these areas, so I'll deep-dive in!
I'll keep you posted and let you know if I need any help along the way .

PR #753, opened 2025-01-14
Trust `$GITHUB_WORKSPACE`  in third-party Workflow

egibs:
Like the other container-based jobs, we need to trust the `$GITHUB_WORKSPACE` (fun aside: `$GITHUB_WORKSPACE` is not the same as `${{ github.workspace }}` when using a container) in the third-party update job (since containers operate differently from VMs).

### DIFF
--- a/.github/workflows/third-party.yaml
+++ b/.github/workflows/third-party.yaml
@@ -36,6 +36,8 @@ jobs:
           apk update
           apk add bash curl findutils gh git go nodejs perl upx xz yara-x-compat
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+      - name: Trust repository
+        run: git config --global --add safe.directory $GITHUB_WORKSPACE
       - uses: chainguard-dev/actions/setup-gitsign@e82b4e5ae10182af72972addcb3fedf7454621c8
       - name: Set up Octo-STS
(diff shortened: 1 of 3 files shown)


.github/workflows/third-party.yaml
```diff
@@ -36,6 +36,8 @@ jobs:
           apk update
           apk add bash curl findutils gh git go nodejs perl upx xz yara-x-compat
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+      - name: Trust repository
+        run: git config --global --add safe.directory $GITHUB_WORKSPACE
```
stevebeattie, line 40: `$GITHUB_WORKSPACE` should probably be quoted; I wouldn't expect there to be a space or other non-ascii in the name, but it's good practice to do so anyway.

stevebeattie (changes requested):
Thanks, one nit.

.github/workflows/third-party.yaml
egibs: Good call. I'll run through all of the references and update them.

egibs: Updated in [`1fbbb0f` (#753)](https://github.com/chainguard-dev/malcontent/pull/753/commits/1fbbb0f9541c20aee5ad3a977e3e1beac39cf18d).

stevebeattie (approved):
Thanks for fixing them up!

PR #755, opened 2025-01-14
Specify bash for PR creation; quote more strings

egibs:
Even though we've installed Bash in the job container, GitHub seems to be using `sh`:
```
sh: third_party/yara/YARAForge/RELEASE: unknown operand
```

This PR explicitly tells the PR creation step to run via `bash` when we create the third-party update PR so that the `[[` test will succeed.

I also went through and quoted remaining string variables for consistency.


stevebeattie (approved):
We don't strictly need the bash form of the test here, but as bash is used elsewhere in the update process, it won't exactly hurt anything either.

Thanks for fixing up more of the variable quoting!

PR #760, opened 2025-01-15
Increase scanner pool safety; fix panics and data races

egibs:
Under certain circumstances (e.g., running in a serverless environment), malcontent will panic when creating too many scanners at one time without any gating or delays.

This PR handles scanner creation and destruction using safer patterns to avoid contention and panics. Performance is as good or better locally (though I was unable to reproduce the panics seen in Cloud Run).


pkg/malcontent/malcontent.go
```diff
@@ -274,7 +306,3 @@ func (p *ScannerPool) Cleanup() {
 	p.scanners = nil
 	atomic.StoreInt32(&p.currentCount, 0)
 }
-
-func (p *ScannerPool) Stats() (int32, int32) {
```
egibs, line 278: We never used this since we can rely on the number of available scanners.

stevebeattie (approved):
Approving, but we may need to revisit further. Thanks!

egibs:
> Approving, but we may need to revisit further. Thanks!

Follow-up changes are in [`dfcdf62` (#760)](https://github.com/chainguard-dev/malcontent/pull/760/commits/dfcdf62776b5ca6afe3bec0e030cb5ea9e7dbe00). Thanks for the feedback!

PR #763, opened 2025-01-15
chore: remove `go` version in golangci config in favor of go version in go.mod

chenrui333:
if not specifying the go version, it would fallback to the version in go.mod, I think that that would be better to keep the go version consistency.


egibs (approved):
Thanks!

PR #765, opened 2025-01-17
Remove scanner pool in favor of single-use scanners

egibs:
The scanner pool concept was neat in theory, but yara-x leverages WASM which exhibits unpredictable behavior when a new scanner is created. Notably, each new scanner is allocated 8GB of virtual memory for the time being and this value is not configurable. Controlling this behavior with a separate value then adds another layer of complexity.

This PR reverts to the original pattern we used for go-yara with the main downside being that we are unable to scan file descriptors with yara-x and must instead read the entire file into memory before scanning it. That said, we'll be able to directly influence how many scanners are created with `c.Concurrency` rather than needing to account for two values. Ultimately, not using the scanner pool has little impact on performance (though I'd much rather have stability right now anyway).

To allow for yara-x to work in environments like Cloud Run, we'll still need a change that allows for WASM tuning to drop the memory reservation to hopefully free up virtual memory overhead and prevent panics.

### DIFF
--- a/cmd/mal/mal.go
+++ b/cmd/mal/mal.go
@@ -63,7 +63,6 @@ var (
 	outputFlag                string
 	profileFlag               bool
 	quantityIncreasesRiskFlag bool
-	scannersFlag              int
 	statsFlag                 bool
 	thirdPartyFlag            bool
 	verboseFlag               bool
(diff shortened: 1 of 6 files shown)


stevebeattie (commented):
https://github.com/chainguard-dev/malcontent/blob/d343832f7337e8311a47e35cf24045f5489aa376/cmd/mal/mal.go#L372-L376
Do want to mark this flag as currently not being honored?

egibs:
> https://github.com/chainguard-dev/malcontent/blob/d343832f7337e8311a47e35cf24045f5489aa376/cmd/mal/mal.go#L372-L376
> 
> Do want to mark this flag as currently not being honored?

Oh, I need to remove that too.

I'm working on a Wolfi PR to help with our reservation problem as in parallel to this.

stevebeattie (approved):

PR #772, opened 2025-01-20
Optimize memory usage in report.go and scan.go

egibs:
This PR leverages the profiling changes from #767 and reduces the memory usage of Malcontent by 1-3GB when scanning many files by using `io.ReadFull` when reading file contents, closing the file immediately after processing, and allowing match strings and the string processor to be GC'd earlier in `report.go`.

Other than that, I made some small safety tweeks re: `mrs` in `Generate`.

I ran these changes several times with `make refresh-sample-testdata` which is a worst case scenario and usage topped out at ~7-8GB instead of 10+. Now that we're using yara-x, we want to work on minimizing our memory footprint as much as possible (although yara-x just shipped a PR to help with virtual address space usage which should be included in the next release).

### DIFF
--- a/pkg/action/scan.go
+++ b/pkg/action/scan.go
@@ -75,7 +75,6 @@ func scanSinglePath(ctx context.Context, c malcontent.Config, path string, ruleF
 	if err != nil {
 		return nil, err
 	}
-	defer f.Close()
 
 	fi, err := f.Stat()
 	if err != nil {
@@ -89,6 +88,7 @@ func scanSinglePath(ctx context.Context, c malcontent.Config, path string, ruleF
 	if err != nil {
 		return nil, err
 	}
+	f.Close()
 
 	mrs, err := yrs.Scan(fc)
 	if err != nil {
@@ -100,7 +100,11 @@ func scanSinglePath(ctx context.Context, c malcontent.Config, path string, ruleF
 	if err != nil {
 		return nil, err
 	}
-	f.Close()
+
+	if fr.Error != "" {
+		logger.Debug("skipping", slog.Any("error", err))
+		return &malcontent.FileReport{Path: path, Error: fmt.Sprintf("generate: %v", fr.Error)}, nil
+	}
 
 	// Clean up the path if scanning an archive
 	var clean string
--- a/pkg/action/scan.go
+++ b/pkg/action/scan.go
@@ -75,6 +75,7 @@ func scanSinglePath(ctx context.Context, c malcontent.Config, path string, ruleF
 	if err != nil {
 		return nil, err
 	}
+	defer f.Close()
 
 	fi, err := f.Stat()
 	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/report/report.go
```diff
@@ -367,7 +367,7 @@ func TrimPrefixes(path string, prefixes []string) string {
 //nolint:cyclop // ignore complexity of 64
 func Generate(ctx context.Context, path string, mrs *yarax.ScanResults, c malcontent.Config, expath string, _ *clog.Logger, fc []byte) (*malcontent.FileReport, error) {
 	if mrs == nil {
-		return nil, fmt.Errorf("yara-x scan results are nil for %s", path)
+		return &malcontent.FileReport{Path: path, Error: fmt.Sprintf("yara-x scan results are nil for %s", path)}, nil
```
egibs, line 370: If `mrs` is somehow `nil` by the time we get here, we just replicate what we do here:
```go
mrs, err := yrs.Scan(fc)
	if err != nil {
		logger.Debug("skipping", slog.Any("error", err))
		return &malcontent.FileReport{Path: path, Error: fmt.Sprintf("scan: %v", err)}, nil
	}
```

pkg/action/scan.go
```diff
@@ -70,11 +71,25 @@ func scanSinglePath(ctx context.Context, c malcontent.Config, path string, ruleF
...
+	}
+
+	size := fi.Size()
+	fc := make([]byte, size)
+
+	_, err = io.ReadFull(f, fc)
+	if err != nil {
+		return nil, err
+	}
+	f.Close()
```
wlynch, line 91:
```suggestion
	f, err := os.Open(path)
	if err != nil {
		return nil, err
	}
	defer f.Close()

	fi, err := f.Stat()
	if err != nil {
		return nil, err
	}

	size := fi.Size()
	fc := make([]byte, size)

	if _, err := io.ReadFull(f, fc); err != nil {
		return nil, err
	}
```

wlynch: Move the close into a defer further up - this way the file will be closed even if you return early from the other error checks.

pkg/action/scan.go
```diff
@@ -86,6 +101,10 @@ func scanSinglePath(ctx context.Context, c malcontent.Config, path string, ruleF
 		return nil, err
 	}
 
+	if fr.Error != "" {
+		return &malcontent.FileReport{Path: path, Error: fmt.Sprintf("generate: %v", fr.Error)}, nil
+	}
+
```
wlynch, line 107: It's a little confusing that errors can be set in `err` or `fr.Error` - can we consolidate these? (e.g. should we have a custom error type that wraps the error message + path?)

egibs: That's a good question. We do this in three other places and it throws me off since I would expect to just return `err` or `fmt.Errorf(...)`.

egibs: I'll fix this up in another PR. We use `fr.Error` to change rendering behavior elsewhere so it's a more systemic change than swapping to a new return pattern.

Really appreciate the suggestion, though 👍🏻

wlynch (approved):

Issue #775, opened 2025-01-21
Less confusing error patterns


egibs:
As pointed out by @wlynch in #772 -- we sometimes return `&malcontent.FileReport{Path: path, Error: ...}` and then check for `fr.Error` rather than using a custom error type that can provide a more idiomatic substitute.

Rip out the `Error` field and add a new custom type instead.

PR #777, opened 2025-01-22
Remove FileReport Error field in favor of custom error type

egibs:
Closes: #775 

We had a couple of instances of leveraging `fr.Error` to influence rendering behavior or take the place of actual errors.

This PR adds a new error type and a few sentinel errors to help with returning more consistent, idiomatic errors. We'll still populate `fr.Skipped` when certain errors are encountered but won't rely on the `Error` field any longer. We still use `fr.Skipped` to ignore rendering problematic files, so nothing changes there.

I also added some more `c.Renderer == nil` checks influenced by #774.

### DIFF
--- a/pkg/action/scan.go
+++ b/pkg/action/scan.go
@@ -36,11 +36,8 @@ var (
 	// compiledRuleCache are a cache of previously compiled rules.
 	compiledRuleCache atomic.Pointer[yarax.Rules]
 	// compileOnce ensures that we compile rules only once even across threads.
-	compileOnce           sync.Once
-	ErrMatchedCondition   = errors.New("matched exit criteria")
-	ErrFileScan           = NewFileReportError(nil, "", "scan failed")
-	ErrGenerateFileReport = NewFileReportError(nil, "", "failed to generate file report")
-	ErrNilFileReport      = NewFileReportError(nil, "", "file report is nil")
+	compileOnce         sync.Once
+	ErrMatchedCondition = errors.New("matched exit criteria")
 )
 
 // scanSinglePath YARA scans a single path and converts it to a fileReport.
@@ -76,31 +73,31 @@ func scanSinglePath(ctx context.Context, c malcontent.Config, path string, ruleF
 
 	f, err := os.Open(path)
 	if err != nil {
-		return nil, NewFileReportError(err, path, errMsgNilReport)
+		return nil, err
 	}
 	defer f.Close()
 
 	fi, err := f.Stat()
 	if err != nil {
-		return nil, NewFileReportError(err, path, errMsgNilReport)
+		return nil, err
 	}
 
 	size := fi.Size()
 	fc := make([]byte, size)
 
 	if _, err := io.ReadFull(f, fc); err != nil {
-		return nil, NewFileReportError(err, path, errMsgNilReport)
+		return nil, err
 	}
 
 	mrs, err := yrs.Scan(fc)
 	if err != nil {
 		logger.Debug("skipping", slog.Any("error", err))
-		return nil, NewFileReportError(err, path, errMsgScanFailed)
+		return nil, err
 	}
 
 	fr, err := report.Generate(ctx, path, mrs, c, archiveRoot, logger, fc)
 	if err != nil {
-		return nil, NewFileReportError(err, path, errMsgGenerateFailed)
+		return nil, NewFileReportError(err, path, TypeGenerateError)
 	}
 
 	// Clean up the path if scanning an archive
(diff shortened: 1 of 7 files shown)


pkg/action/scan.go
```diff
@@ -36,8 +36,11 @@ var (
 	// compiledRuleCache are a cache of previously compiled rules.
 	compiledRuleCache atomic.Pointer[yarax.Rules]
 	// compileOnce ensures that we compile rules only once even across threads.
-	compileOnce         sync.Once
-	ErrMatchedCondition = errors.New("matched exit criteria")
+	compileOnce           sync.Once
+	ErrMatchedCondition   = errors.New("matched exit criteria")
+	ErrFileScan           = NewFileReportError(nil, "", "scan failed")
+	ErrGenerateFileReport = NewFileReportError(nil, "", "failed to generate file report")
+	ErrNilFileReport      = NewFileReportError(nil, "", "file report is nil")
```
wlynch, line 43: Is this an actual error? 🤔 Mostly asking because we have this later:

```
case TypeNilError:
		return nil, nil
```

wlynch, line 43: Move these into scan_error.go and use the consts to make sure that these can't diverge.

wlynch: Actually, are these used anywhere?

pkg/action/scan_error.go
```diff
@@ -0,0 +1,67 @@
...
+	TypeNilError
+)
+
+type FileReportError struct {
+	err    error
+	path   string
+	reason string
+}
+
+func NewFileReportError(err error, path, reason string) *FileReportError {
```
wlynch, line 25:
```suggestion
func NewFileReportError(err error, path, reason ErrorType) *FileReportError {
```

Using a freeform string here makes it harder for external packages to use the typing effectively since they need to be exact string matches for the type switch to work. Consider using the ErrorTypes enum directly here, and converting them to strings in Error().

pkg/action/scan_error.go
```diff
@@ -0,0 +1,67 @@
...
+	errMsgGenerateFailed = "failed to generate file report"
+	errMsgNilReport      = "file report is nil"
+)
+
+type ErrorType int
+
+const (
+	TypeScanError ErrorType = iota
+	TypeGenerateError
+	TypeNilError
```
wlynch, line 16: Godoc comments please 🙏

pkg/action/scan_error.go
wlynch: If you do this though, add a new TypeUnknown enum so that 0 defaults to Unknown rather than TypeScanError

pkg/action/scan.go
egibs: I may have duped while moving things around. I'll clean these up 👍🏻

egibs: It's technically not an error. We can actually rely on the existing `fr == nil` check to return `nil, nil`.

wlynch: 👍 sgtm. document this behavior in the method's godoc! 🙏

pkg/action/scan_error.go
egibs: Added in [`8e1e5c6` (#777)](https://github.com/chainguard-dev/malcontent/pull/777/commits/8e1e5c6cf378a547f46bccdd7d73ceb9a912eb5b). Let me know if you think any other comments are needed!

pkg/action/scan_error.go
egibs: Added in [`8e1e5c6` (#777)](https://github.com/chainguard-dev/malcontent/pull/777/commits/8e1e5c6cf378a547f46bccdd7d73ceb9a912eb5b).

pkg/action/scan.go
egibs: Updated in [`8e1e5c6` (#777)](https://github.com/chainguard-dev/malcontent/pull/777/commits/8e1e5c6cf378a547f46bccdd7d73ceb9a912eb5b).

egibs: Updated in [`8e1e5c6` (#777)](https://github.com/chainguard-dev/malcontent/pull/777/commits/8e1e5c6cf378a547f46bccdd7d73ceb9a912eb5b).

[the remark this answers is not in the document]
egibs: Added in [`8e1e5c6` (#777)](https://github.com/chainguard-dev/malcontent/pull/777/commits/8e1e5c6cf378a547f46bccdd7d73ceb9a912eb5b).

pkg/action/scan_error.go
```diff
@@ -0,0 +1,80 @@
...
+
+// Error type iotas.
+// TypeUnknown will be the default of `0`.
+// TypeScanError is to be used when compiled rules are invalid or the scan fails otherwise.
+// TypeGenerateError is to be used when a file's report cannot be created.
+const (
+	TypeUnknown ErrorType = iota
+	TypeScanError
+	TypeGenerateError
+)
```
wlynch, line 24:
```suggestion
const (
	// TypeUnknown will be the default of `0`.
	TypeUnknown ErrorType = iota
	// TypeScanError is to be used when compiled rules are invalid or the scan fails otherwise.
	TypeScanError
	// TypeGenerateError is to be used when a file's report cannot be created.
	TypeGenerateError
)
```

Needs to be inlined for godoc to know how to map them to each type.

wlynch (approved):
one last nit, otherwise lgtm! Thanks.

pkg/action/scan_error.go
egibs: Excellent -- thanks!

wlynch (approved):

PR #779, opened 2025-01-25
Extract .jar and .zip files concurrently, use buffer for all io.Copy operations

egibs:
We can speed up the extraction of large packages like Spark and Trino by using concurrency since both contain many `.jar` files. Both `.jar` and `.zip` archives can be extracted in parallel rather than sequentially which is fortuitous in this case since these packages, when fully extracted, amount to somewhere in the neighborhood of ~1e5 files.

Additionally, we can improve overall memory usage by using a buffer pool and `io.CopyBuffer` across all extraction methods.

Finally, I fixed `.xz` extractions not using a limit reader which slipped through previous optimizations.


stevebeattie (approved):
LGTM, and saw a slight performance increase in limited testing. Thanks!

PR #823, opened 2025-03-11
action: Pull interactive check out into a function

jonjohnsonjr:
As per #817 this pattern is a bit confusing an error prone, so let's extract it into a function.


egibs (approved):
Thanks for cleaning this up 🙂

PR #842, opened 2025-03-25
feat: add basic zstd compression support

stevebeattie:
Malcontent was not properly scanning zstd compressed files e.g. kernel modules on modern Ubuntu systems. As an example, without this change:

```
$  mal --format=simple --verbose analyze /lib/modules/6.11.0-19-generic/kernel/fs/smb/server/ksmbd.ko.zst
time=2025-03-24T20:51:36.262-07:00 level=DEBUG source=$HOME/git/chainguard-dev/malcontent/pkg/action/scan.go:71 msg="skipping /usr/lib/modules/6.11.0-19-generic/kernel/fs/smb/server/ksmbd.ko.zst [<unknown>]: data file or empty" path=/usr/lib/modules/6.11.0-19-generic/kernel/fs/smb/server/ksmbd.ko.zst
```

With this patch applied:

```
$ ./mal --format=simple --verbose analyze /lib/modules/6.11.0-19-generic/kernel/fs/smb/server/ksmbd.ko.zst
time=2025-03-24T20:53:47.375-07:00 level=DEBUG source=$HOME/git/chainguard-dev/malcontent/pkg/archive/archive.go:110 msg="creating temp dir" path=/usr/lib/modules/6.11.0-19-generic/kernel/fs/smb/server/ksmbd.ko.zst
time=2025-03-24T20:53:47.375-07:00 level=DEBUG source=$HOME/git/chainguard-dev/malcontent/pkg/archive/zstd.go:18 msg="extracting zstd" dir=$HOME/tmp/ksmbd.ko.zst439390431 file=/usr/lib/modules/6.11.0-19-generic/kernel/fs/smb/server/ksmbd.ko.zst
c2/addr/ip: medium
crypto/aes: low
crypto/cipher: medium
fs/attributes/remove: medium
fs/attributes/set: medium
fs/directory/create: low
fs/directory/remove: low
fs/file/delete: low
fs/file/open: low
fs/lock_update: low
impact/remote_access/heartbeat: medium
net/ip/send_unicast: low
net/rpc/ntlm: medium
net/socket/listen: medium
net/socket/peer_address: low
net/socket/receive: low
net/socket/send: low
os/kernel/netlink: low
persist/daemon: medium
persist/kernel_module/module: medium
persist/kernel_module/name: medium
sus/exclamation: medium
```

This patch was mostly copy-wasting from the bz2 archive implementation and cherry-picking bits and bobs from the zstd support in the rpm.go implementation.


stevebeattie:
A couple of more broad thoughts:

- these single file compressed formats (bz2, gz, zstd, xz) feel like they could be abstracted into a common generic method that takes a data structure consisting of the type, standard suffixes, and decompression function as an argument, to extract out all the common file handling code into one place, so that it doesn't need to copy-pasta'd all over the place. (I didn't check whether the gz and xz implementations support an `io.Reader` interface.)

- some simple test cases for all the different archive types would be nice.

egibs:
Method looks good to me -- can you add `zst` and `zstd` here?
https://github.com/chainguard-dev/malcontent/blob/5254e43c7f21ecc63470d7599f54c69c2ae09c60/pkg/archive/archive.go#L202-L222

That function kind of centralizes the logic but we still fan out to the specific extraction code to keep the code somewhat readable.

stevebeattie:
> Method looks good to me -- can you add `zst` and `zstd` here?

Doh, I had made that change locally, but missed adding it to the commit; the dangers of late night hacking. Added and pushed.

PR #867, opened 2025-04-19
Remove extracted files after scanning, add sync.Pools, other optimizations/cleanup

egibs:
In order to reduce the memory footprint of scans, particularly in cases like Cloud Run (with its in-memory filesystem) or `/tmp` being `tmpfs`, we can immediately clean up skipped and processed files within archives. Since fully-extracted archives can contain many thousands of files, there's no need to leave these around once scanned.

I also added sync.Pools for scanners and buffers/byte slices that should all be thread-safe in the same way our compiled rules are. The new pools should keep memory pressure lower since they can be re-used throughout the lifecycle of a scan.

Finally, I went through and made some modernization/optimization tweaks. The largest of these changes is filtering `c.Scan` runs prior to report generation to shorten the loop for files below the minimum risk threshold (3) wherein we avoid calculating checksums and initializing a file report that won't be used aside from specifying the skipped reason. Some of the other changes clean up needlessly complex code which had no performance benefit (`strings.go` in the report package is a good example).


pkg/action/scan.go
```diff
@@ -83,20 +106,34 @@ func scanSinglePath(ctx context.Context, c malcontent.Config, path string, ruleF
...
+	defer filePool.Put(fc)
 	if _, err := io.ReadFull(f, fc); err != nil {
 		return nil, err
 	}
 
-	mrs, err := yrs.Scan(fc)
+	// Immediately remove archive files read into memory
+	if isArchive {
+		os.RemoveAll(path)
+	}
```
wlynch, line 119: Should this be in a defer so we can guarantee this happens even if we encounter an err somewhere above?

pkg/archive/archive.go
```diff
@@ -11,30 +11,105 @@ import (
...
+	initializeOnce                sync.Once
+)
 
 // isValidPath checks if the target file is within the given directory.
 func IsValidPath(target, dir string) bool {
 	return strings.HasPrefix(filepath.Clean(target), filepath.Clean(dir))
 }
 
-// ExtractArchiveToTempDir creates a temporary directory and extracts the archive file for scanning.
+func extractNestedArchive(ctx context.Context, d string, f string, extracted *sync.Map) error {
```
egibs, line 33: This is the original function (pre-#864) with a few tweaks.

pkg/archive/archive.go
```diff
@@ -11,30 +11,105 @@ import (
...
+			if err := extractNestedArchive(ctx, d, rel, extracted); err != nil {
+				return fmt.Errorf("process nested file %s: %w", rel, err)
+			}
+		}
+	}
+	return nil
+}
+
+// extractArchiveToTempDir creates a temporary directory and extracts the archive file for scanning.
 func ExtractArchiveToTempDir(ctx context.Context, path string) (string, error) {
```
egibs, line 113: This is the original function (pre-#864) with the pool tweaks.

Makefile
```diff
@@ -70,11 +70,11 @@ $(YARA_X_BIN):
 
 LINTERS += golangci-lint-lint
 golangci-lint-lint: $(GOLANGCI_LINT_BIN)
-	find . -name go.mod -execdir "$(GOLANGCI_LINT_BIN)" run -c "$(GOLANGCI_LINT_CONFIG)" \;
+	find . -maxdepth 1 -name go.mod -execdir "$(GOLANGCI_LINT_BIN)" run -c "$(GOLANGCI_LINT_CONFIG)" \;
```
egibs, line 73: This prevents the `yara-x` Go code from being linted in `out/`.

wlynch (approved):

pkg/action/scan.go
egibs: Added in [`cf0b6ca` (#867)](https://github.com/chainguard-dev/malcontent/pull/867/commits/cf0b6ca496d50982cf0d09b765fab4ef53437168).

pkg/programkind/programkind.go
```diff
@@ -105,22 +106,10 @@ type FileType struct {
 	MIME string
 }
 
-var fileTypeCache = sync.Map{}
-
-// GetCachedFileType returns previously-seen filetypes or stores novel filetypes for future retrieval
-func GetCachedFileType(path string) (*FileType, error) {
```
egibs, line 111: This was the opposite of memory-conscious and didn't really net us anything.

Issue #696, opened 2024-12-11
Consider "filetype" metadata when executing rules


tstromberg:
For example, .py files should only run rules without a filetype defined or rules with "py" in the file types.

One weird idea I had was precompiling groups of rules based on `programkind.supportedKind`, so for example, rules[py] would have all of the precompiled rules for python files.

Related to #454

tstromberg:
While this issue talks about only executing rules for particular file types, with all the performance improvements we've landed over the last 6 months, it'd be sufficient to just filter out results that don't match the filetypes set by a rule.

This would allow us to remove false-positives from Python-specific detections that matching random JSON files or Go binaries.

tstromberg:
While this issue initially mentioned a `filetype` metadata tag, after adding it to rules, I think a comma-separated `filetypes` makes more sense: `filetypes: "py,js,elf,macho"`

Issue #454, opened 2024-09-15
Add "filetypes" metadata to rules


tstromberg:
To improve performance and reduce false positives, I'd like to add metadata to our rules to constrain which file types they are associated with. For example:

```
meta:
   filetypes = "python,javascript"
```

I expect that most rules will be associated with a single file type. As far as standardization goes, I recommend basing it loosely on MIME types, specifically the second half of it. Looking at https://www.sitepoint.com/mime-types-complete-list/ - I would say our filetypes would look something like:

- python
- javascript
- shell
- c

For examples where the mime types don't exist or are excessively long, we can make up something meaningful, like:

- macho
- elf

tstromberg:
cc @egibs for his thoughts.

This is related to #382

egibs:
I'm all for this. The hard part will be going through each rule and adding the metadata and we still don't have a way to layer customizations on top of third party rules so TBD how complete our coverage will end up being.

tstromberg:
I’m considering third party rules out of scope for this at the moment.
There may be however some interesting heuristics we can do to autogenerate
this metadata based on tags or other metadata fields supported by 3rd party
rules.

But yes, adding metadata to each rule will be painful, but I’m down to do
it as part of a general cleanup.

tstromberg:
A number of rules have this now: I think the next step will be updating our engine to respect this file type. Then we will be more likely to implement it everywhere. 

For example, .py files should only run rules without a filetype defined, or those with "py" in the file types.

PR #898, opened 2025-05-01
Add file type support for report generation

egibs:
Closes: #696 
Relates to: #454 

This PR finally adds filetype support for rules. To better support this, I also added `filetypes` to rules that seemed to be obviously referencing an OS/language/etc.

This also allows us to remove the `probably` rules that have to be replicated and updated separately (and they don't really work all that well anyway) -- as an aside, `yara-x` 0.15.0 added support for `include` which means that we can define global rules centrally and then reference them in the appropriate rules but that's for a future PR.

If the `filetypes` metadata key is present, then we only count the rule as a match if the file's extension is one of the comma-delimited strings and any number of extensions can be added to the metadata value. Finally, rules without the `filetypes` key are considered universal.


pkg/action/testdata/scan_archive
```diff
@@ -1070,15 +1081,15 @@
                     "RuleName": "go_file_read"
                 },
                 {
-                    "Description": "renames files",
+                    "Description": "rename",
```
tstromberg, line 1084: this description change seems off?

rules/anti-behavior/random_behavior.yara
```diff
@@ -43,6 +45,7 @@ rule java_random: low {
 rule rand_call: medium {
   meta:
     description = "exhibits random behavior"
+    filetypes   = "text/x-c,text/x-php,text/x-perl"
```
tstromberg, line 48: With verbose filetypes like these, it's less likely that folks will willingly update this list for rules. Can we use the file extension associated to the MIME type instead? Such as: `filetypes = "c,php,p"` for this example?

Most folks don't have MIME types memorized, whereas they are likely to have file extensions memorized. For example, `.exe` vs [application/vnd.microsoft.portable-executable](https://www.iana.org/assignments/media-types/application/vnd.microsoft.portable-executable)

Even if someone has a MIME type memorized, there is often unclear standards on which to use. For example, for `.zip` there are 3 MIME types in use :(

tests/linux/clean/kibana/ac96ceb8-4399-4191-af1d-4feeac1f1f46_108.json.simple
```diff
@@ -0,0 +1,11 @@
+# linux/clean/kibana/ac96ceb8-4399-4191-af1d-4feeac1f1f46_108.json: medium
```
tstromberg, line 1: I think we had a JSON regression here; We previously skipped JSON files unless they matched a particular pattern indicative of NPM packaging:

https://github.com/chainguard-dev/malcontent/blob/8c381b30bd77d222e8e3e434c908670db41c2fea/pkg/programkind/programkind.go#L195

tstromberg (changes requested):

pkg/action/testdata/scan_archive
egibs: I think it matched a different rule with the changes. I can double-check.

rules/anti-behavior/random_behavior.yara
egibs: Good point. This was the easiest since the map was set up that way, but perhaps we can flip that around and make this cleaner.

tests/linux/clean/kibana/ac96ceb8-4399-4191-af1d-4feeac1f1f46_108.json.simple
egibs: Reverted in [`81c9d1f` (#898)](https://github.com/chainguard-dev/malcontent/pull/898/commits/81c9d1f8c1f62e1728da01a104331e92dbd24af7).

rules/anti-behavior/random_behavior.yara
egibs: Swapped over to referencing extensions in [`24b9bf2` (#898)](https://github.com/chainguard-dev/malcontent/pull/898/commits/24b9bf2d1a69e487c5a5df8fcb33fcbf022be7eb).

pkg/action/testdata/scan_archive
egibs: Fixed in [`a51618e` (#898)](https://github.com/chainguard-dev/malcontent/pull/898/commits/a51618ef6aa1a382d92131da3a533c74fd5b7403) and [`2e0fd4a` (#898)](https://github.com/chainguard-dev/malcontent/pull/898/commits/2e0fd4a37854f78eaab9fc4b3c33f0c5e0131410).

eslerm (approved):
:sunglasses: cheers

PR #904, opened 2025-05-07
fix: Reduce Malcontent JS False Positives

antitree:
This fix adjusts the severity on some of the obfuscation rules and one of the rules for javascript reversing that was producing a high number of false positives. I think we can roll this change back after we've tuned these rules to producer higher quality results.


eslerm (approved):
Thank you for jumping on this!

eslerm (approved):
cheers for test fix

PR #924, opened 2025-05-14
Allow echo_decode_bash* to execute on larger files

tstromberg:
I noticed this detection didn't fire in the `tj-actions` incident, due to it being within a 2MB JS file.


eslerm (approved):
It would be interesting to test alerts with and without this change.

PR #925, opened 2025-05-14
Improve Javascript results, particularly for rand-user-agent

tstromberg:
Updates based on analysis done on the artifacts within https://www.aikido.dev/blog/catching-a-rat-remote-access-trojian-rand-user-agent-supply-chain-compromise


eslerm (approved):
Thanks! I'd really like to do an archive malcontent audit based on PRs :)

PR #937, opened 2025-05-20
Use the `include` directive to centralize common private rules

egibs:
`yara-x` 0.15.0 introduced the `include` directive in parity with Yara. This PR leverages the directive by storing commonly-used private rules in a central `globals.yara` file which can then be included in rule files as necessary. 

This directive inserts the included file's contents in each file but since we namespace by path we can have any number of duplicate rule names without encountering compilation errors.


pkg/compile/compile.go
```diff
@@ -159,16 +166,57 @@ func removeRules(data []byte, rulesToRemove []string) []byte {
...
+	return modified
+}
+
 func Recursive(ctx context.Context, fss []fs.FS) (*yarax.Rules, error) {
 	if ctx.Err() != nil {
 		return nil, ctx.Err()
 	}
 
-	yxc, err := yarax.NewCompiler(yarax.ConditionOptimization(true))
+	yxc, err := yarax.NewCompiler(yarax.ConditionOptimization(true), yarax.EnableIncludes(true))
```
egibs, line 203: `yarax.EnableIncludes` is true by default, but I wanted to have it here explicitly for reference.

pkg/compile/compile.go
```diff
@@ -177,14 +225,21 @@ func Recursive(ctx context.Context, fss []fs.FS) (*yarax.Rules, error) {
 				return err
 			}
 
-			if !d.IsDir() && (filepath.Ext(path) == ".yara" || filepath.Ext(path) == ".yar") {
```
eslerm, line 180: cheers for reducing complexity

.github/workflows/style.yaml
```diff
@@ -31,14 +31,14 @@ jobs:
 
       - name: Install yara-x
         run: |
-          wget https://github.com/VirusTotal/yara-x/releases/download/v0.10.0/yara-x-v0.10.0-x86_64-unknown-linux-gnu.gzip -O yara-x.gzip
```
eslerm, line 34: This is a large version bump v0.10.0 was released in Nov 26, 2024 a v0.15.0 was released yesterday.

Is it necessary to include with the rest of the PR? Not a block, just curious.

rules/anti-static/elf/entropy.yara
```diff
@@ -1,22 +1,14 @@
-import "math"
```
eslerm, line 1: If I were to paint this shed, I would always place `import` statements before `include`.

YARA allows either order, but I prefer following Python's style and it makes some sense for `include` to be next to the other rules.

Not a block, just a nit :)

Appreciate that variable names are being converted to snake case too.

.github/workflows/style.yaml
eslerm: re-read the description!

egibs: Ah, yea, this should have been handled with each new version update but it's not tied to the Makefile so it drifted. We should probably fix this in a future PR so that we don't have to manage this separately.

rules/anti-static/elf/entropy.yara
egibs: No disagreements here. `include` after `import` is technically more correct since it's adding the included file's contents to the file which would follow `import` if it were added directly to the file to begin with.

Updated in: [`cb0a939` (#937)](https://github.com/chainguard-dev/malcontent/pull/937/commits/cb0a939dd04b1a3be0d630f053b5b9afdcb7b6a6).

.github/workflows/style.yaml
eslerm: Cheers, made an issue for that.

eslerm (approved):

PR #941, opened 2025-05-21
feat: purego package update

daeho-ro:
There is a build error on macos 15 and need to update purego.

 https://github.com/golang/go/issues/73617


egibs:
Thanks for the PR! I assume this came up as part of https://github.com/Homebrew/homebrew-core/pull/223227?

We pull this in as part of `gopsutil` but it sounds like it supports user-defined versions of `purego` so this should be safe.

egibs (approved):

PR #944, opened 2025-05-21
Support absolute `include` paths when calling malcontent externally

egibs:
The changes from #937 to replace the relative include path with its absolute path worked in the context of this repository's directory structure but not when installed as a module and used elsewhere (e.g., importing malcontent as a library).

This PR should allow us to locate `rules/` regardless of where we call malcontent from. I've validated this where I saw the failure and it works with these changes.

### DIFF
--- a/pkg/compile/compile.go
+++ b/pkg/compile/compile.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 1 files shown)


tstromberg (approved):

eslerm (commented):
Thanks for clarifying variables names while adding this check

PR #949, opened 2025-05-22
Reduce FPs for common JS files

egibs:
This PR tweaks the `fromChar` rules to ignore common JS files that we see when scanning packages.

Most of the failures are from `.js` and `.ts` files while some where from unrelated filetypes which will be addressed by #898.

### DIFF
--- a/rules/exec/remote_commands/code_eval.yara
+++ b/rules/exec/remote_commands/code_eval.yara
@@ -50,62 +50,30 @@ rule js_eval_response: critical {
     filesize < 1MB and any of ($val*)
 }
 
-rule js_eval_near_enough_fromChar: high {
+rule js_eval_near_enough_fromChar: medium {
   meta:
-    description = "Likely executes encrypted content"
+    description = "Evaluates content via String.fromCharCode"
     filetypes   = "js,ts"
 
   strings:
-    $exec    = /[\s\{]eval\(/
+    $eval    = /[\s\{]eval\(/
     $decrypt = "String.fromCharCode"
 
-    $not_code_server = "fromCharCode(...codes: number[]): string;"
-    $not_grafana     = "self.webpackChunkgrafana=self.webpackChunkgrafana||[]).push("
-    $not_jupyter     = "self[\" webpackChunk_jupyterlab_application_top \"]=self[\" webpackChunk_jupyterlab_application_top \"]||[]).push("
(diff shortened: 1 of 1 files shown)


rules/exec/remote_commands/code_eval.yara
```diff
@@ -59,8 +59,24 @@ rule js_eval_near_enough_fromChar: high {
     $exec    = /[\s\{]eval\(/
     $decrypt = "String.fromCharCode"
 
+    $not_code_server = "fromCharCode(...codes: number[]): string;"
+    $not_grafana     = "self.webpackChunkgrafana=self.webpackChunkgrafana||[]).push("
+    $not_jupyter     = "self[\" webpackChunk_jupyterlab_application_top \"]=self[\" webpackChunk_jupyterlab_application_top \"]||[]).push("
+    $not_jupyter2    = "self[\" webpackChunk_JUPYTERLAB_CORE_OUTPUT \"] = self[\" webpackChunk_JUPYTERLAB_CORE_OUTPUT \"] || []).push([[132,7061]"
+    $not_jupyter3    = "self[\" webpackChunk_jupyterlab_application_top \"]=self[\" webpackChunk_jupyterlab_application_top \"]||[]).push([[227],{69119:"
+    $not_jupyter4    = "self[\" webpackChunk_jupyterlab_application_top \"]=self[\" webpackChunk_jupyterlab_application_top \"]||[]).push([[9296],{49296:"
+    $not_jupyter5    = "self[\" webpackChunk_jupyterlab_application_top \"]=self[\" webpackChunk_jupyterlab_application_top \"]||[]).push([[4470],{27902:"
```
tstromberg, line 68: Since this is probably triggering on webpack, can we just use `webpackChunk` as a signal instead of these other jupyter ones?

tstromberg: On second thought, with this many falses, I'd downgrade this rule to a medium.

tstromberg (changes requested):

rules/exec/remote_commands/code_eval.yara
egibs: Yea, dropping the severities to medium and high respectively would make a huge difference.

egibs: Updated in [`edf3d67` (#949)](https://github.com/chainguard-dev/malcontent/pull/949/commits/edf3d6766da6907f84f71a873df740019f04a81d).

eslerm (approved):

PR #967, opened 2025-05-29
Address extraction edge cases re: duplicate file names

egibs:
More fallout from scanning Sonarqube.

There are five files of interest in this package:
```
./linux-x64-musl/node.xz
./linux-x64/node.xz
./win-x64/node.exe.xz
./darwin-arm64/node.xz
./sonarjs-1.0.0.tgz
```

Our original extraction logic would only process one of the extracted `node` files because only the file name (`node`) was stored in the extracted file map. Additionally, extracting the files would fail with legitimate EOF errors and this was also happening with the `sonarjs-1.0.0.tgz` file. There was also an opportunity to clean up the awkward `extractedFiles` map ranging and replace it with `filepath.WalkDir`.

This PR addresses this edge case, decompresses files in such a way to preserve their parent directory, and removes archive types from `supportedKind` which causes weird behavior. Instead, we can return a non-nil `FileType` if a given path's extension is present in the archive map.

The test files being erased is expected; these files are actually `tar` files for some reason (?) and removing `tar` from `supportedKind` means they aren't scanned directly.


eslerm (approved):
appreciate simplifications :)

PR #968, opened 2025-05-29
Migrate tablewriter from 0.0.5 to 1.0.7

egibs:
We've been getting Dependabot PRs to update `tablewriter` to the new stable version.

`v0.0.5` is a "legacy" version now. This PR updates our usage in accordance with the new v1.x.x methods: https://github.com/olekukonko/tablewriter/blob/fbb970f655f4eaff4aea6ed7830a4c6ad99b1ae7/MIGRATION.md?plain=1#L423

This was a breaking change so Dependabot couldn't handle it automatically.


eslerm (approved):
Thanks for pruning and replacing a legacy dependency \o/

PR #972, opened 2025-05-30
tests: generate coverage info report

stevebeattie:
Adds two toplevel make targets:

- `make coverage-html`: generates out/coverage.html
- `make coverage-browser`: opens a browser page with the coverage report loaded

There are additional support make targets.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -27,6 +27,16 @@ integration tests:
 
 ```make integration```
 
+## Viewing test coverage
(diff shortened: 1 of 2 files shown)


stevebeattie:
Sample output of the coverage information report, as generated against the current tip of `main` (gzipped because github won't allow the html to be uploaded directly):

[coverage.html.gz](https://github.com/user-attachments/files/20525687/coverage.html.gz)

PR #976, opened 2025-06-02
2025/06/02 false positive reduction

egibs:
This PR addresses most of the outstanding false positives we've seen recently. The plan is to merge this and run another audit. Any outstanding findings can be picked up in a subsequent PR.

### DIFF
--- a/rules/anti-static/obfuscation/bitwise.yara
+++ b/rules/anti-static/obfuscation/bitwise.yara
@@ -173,7 +173,7 @@ rule bidirectional_bitwise_math_php: high {
     filesize < 192KB and all of them
 }
 
-rule bitwise_obfuscation: critical {
+rule bitwise_obfuscation: high {
   meta:
     description = "uses bitwise math to obfuscate code"
     ref         = "https://www.reversinglabs.com/blog/python-downloader-highlights-noise-problem-in-open-source-threat-detection"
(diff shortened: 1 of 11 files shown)


rules/anti-static/obfuscation/bitwise.yara
```diff
@@ -188,6 +188,17 @@ rule bitwise_obfuscation: critical {
     $f_substr  = "substr("
     $f_ord     = "ord("
 
+    $not_phpseclib1 = "http://phpseclib.sourceforge.net"
```
antitree, line 191: Do share my perspective that this might be a losing strategy no matter what when we're trying to catch obfuscation as a signal? Bitwise operations and paddings seem really hard to key on and we'll have to keep updating exceptions. 

For me I'd like some kind of confidence score for each rule that gets fed into the decision around blocking/no blocking and these exceptions could improve our confidence score. 

This for example is an item that by itself would be low confidence but mixed with other signal might result in something higher. 

Short term: I feel this is high noise and therefore can't be a blocker. 
Long term: Something that will allow us to tweak the implementation of the rulesets for our environment, without the criticality score.

tests/npm/2024.testerrrrrrrrrr/init.js.simple
```diff
@@ -2,7 +2,7 @@
 anti-static/obfuscation/bool: medium
 anti-static/obfuscation/hex: medium
 anti-static/obfuscation/js: high
-anti-static/obfuscation/url: critical
+anti-static/obfuscation/url: high
```
antitree, line 5: :+1:

antitree (changes requested):
I think we can approve this in the short term but I'm interested in your response to the comments. If the goala are 1) protect the product 2) do not block a team on a false positive we're doing a good job at 1 but what's our assessment of 2?

eslerm (approved):

PR #985, opened 2025-06-04
Ignore Python comments within two rules

egibs:
Two new findings (from a new package PR). Both of these are in comments so we can ignore them easily.

### DIFF
--- a/rules/exec/install_additional/pip_install.yara
+++ b/rules/exec/install_additional/pip_install.yara
@@ -57,8 +57,12 @@ rule pip_installer_url: critical {
   strings:
     $ref = /pip.{1,5}install.{1,4}https{0,1}:\/\/.{0,64}/
 
+    $not_langchain_comment1 = "Please install the exllamav2 library with (cuda 12.1 is required)"
+    $not_langchain_comment2 = "example : "
+    $not_langchain_comment3 = "\"!python -m pip install https://github.com/turboderp/exllamav2/releases/download/v0.0.12/exllamav2-0.0.12+cu121-cp311-cp311-linux_x86_64.whl\""
+
   condition:
-    filesize < 8192 and $ref
+    filesize < 8192 and $ref and none of ($not*)
 }
 
 rule pip_installer_socket: critical {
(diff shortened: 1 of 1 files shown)


rules/exec/install_additional/pip_install.yara
```diff
@@ -57,8 +57,12 @@ rule pip_installer_url: critical {
   strings:
     $ref = /pip.{1,5}install.{1,4}https{0,1}:\/\/.{0,64}/
 
+    $not_langchain_comment1 = "Please install the exllamav2 library with (cuda 12.1 is required)"
+    $not_langchain_comment2 = "example : "
+    $not_langchain_comment3 = "\"!python -m pip install https://github.com/turboderp/exllamav2/releases/download/v0.0.12/exllamav2-0.0.12+cu121-cp311-cp311-linux_x86_64.whl\""
```
eslerm, line 62: These seem awfully specific

rules/exec/install_additional/pip_install.yara
```diff
@@ -69,8 +73,10 @@ rule pip_installer_socket: critical {
   strings:
     $ref = /pip.{1,5}install.{1,4}socket/
 
+    $not_langchain_comment1 = "\"Please install it with `pip install websocket-client`.\""
```
eslerm, line 76: Does this variable need to be overloaded?

eslerm (approved):
Cheers. Approved, but please see comments.

rules/exec/install_additional/pip_install.yara
egibs: Yep, this is pretty ideal. We generally want to exclude strings without being overly-vague to avoid capturing other findings.

rules/exec/install_additional/pip_install.yara
egibs: Not necessarily but this captures the comment as a whole:
https://github.com/langchain-ai/langchain-community/blob/7de6c0dd71c0b456a1473631271b0f92bb0f0925/libs/community/langchain_community/llms/exllamav2.py#L89

PR #992, opened 2025-06-06
Clean up Generate function in report.go

### LINKED ISSUE #751: Refactor `Generate` in `report.go`
Similar to #497 -- `Generate` in `report.go` has grown to be a very sizable function. Much of its functionality can be moved to smaller, more modular functions.

Ritish134:
Broken down the functionality of the generate function into modular functions:
 `buildIgnoreMap()` — creates the ignore tag map

`trimDisplayPath()` — handles trimming file paths

`initFileReport()` — initializes the report struct

`createMatchRulesMap()` —  Store match rules in a map

`processMatchedStrings()` — processes all pattern matches

`buildBehavior()` — creates a behavior object from rule data

`handleMetadata()` — processes metadata and applies overrides

`updateBehavior()` — updates behavior entries in the file report

fixes #751


pkg/report/report.go
```diff
@@ -465,105 +445,11 @@ func Generate(ctx context.Context, path string, mrs *yarax.ScanResults, c malcon
...
-			case "pledge":
-				pledges = append(pledges, v)
-			case "syscall":
-				sy := strings.Split(v, ",")
-				syscalls = append(syscalls, sy...)
-			case "cap":
-				caps = append(caps, v)
-			}
-		}
+		handleMetadata(m, b, fr, override, mrsMap, pledges, caps, syscalls)
```
egibs, line 452: I noticed we're missing pledges and syscalls (and most likely caps even though we don't currently have any samples with this) in the recent test diffs.

We can probably move `pledges`, `caps`, and `syscalls` to `handleMetadata`'s scope.

To verify your changes locally you can run `make refresh-sample-testdata` to build the latest changes and refresh our sample data. If you see any diffs, you can quickly iterate and try again rather than waiting for CI.

egibs:
@Ritish134 -- I made a few conflicting changes in #1017 -- namely calling `processor.clearFileContent()` which will need to be adapted to your new `processMatchedStrings` method.

Once that's fixed, I think this will be in good shape.

Ritish134:
@egibs Done

PR #1000, opened 2025-06-12
Add exception for php_image_include rule

egibs:
This PR adds a straightforward exception for the embedding of Symfony's favicon as base64 in the `php_image_include` rule.

### DIFF
--- a/rules/evasion/covert-location/php_image.yara
+++ b/rules/evasion/covert-location/php_image.yara
@@ -7,8 +7,12 @@ rule php_image_include: critical {
     $php     = "<?php"
     $include = /include\s*\(\s*[^\.]+\.(png|jpg|gif|bmp)/
 
+    // https://github.com/symfony/symfony/blob/7.4/src/Symfony/Component/ErrorHandler/Resources/views/exception_full.html.php#L9
+    // https://github.com/symfony/symfony/blob/7.4/src/Symfony/Component/ErrorHandler/Resources/assets/images/favicon.png.base64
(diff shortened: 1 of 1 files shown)


eslerm (approved):

rules/evasion/covert-location/php_image.yara
```diff
@@ -7,8 +7,13 @@ rule php_image_include: critical {
...
     $include = /include\s*\(\s*[^\.]+\.(png|jpg|gif|bmp)/
 
+    // https://github.com/symfony/symfony/blob/7.4/src/Symfony/Component/ErrorHandler/Resources/views/exception_full.html.php#L9
+    // https://github.com/symfony/symfony/blob/7.4/src/Symfony/Component/ErrorHandler/Resources/assets/images/favicon.png.base64
+    $not_symfony1 = "include('assets/images/favicon.png.base64')"
+    $not_symfony2 = "<a href=\"https://symfony.com/doc/<?= Symfony\\Component\\HttpKernel\\Kernel::VERSION; ?>/index.html\">"
+
   condition:
-    filesize < 5242880 and all of them
+    filesize < 5242880 and all of them and none of ($not*)
```
stevebeattie, line 16: I've felt like I should ask this about other false positive mitigations we've been putting in place, but should the logic here be instead:

`and not all of $(not*)` as in, ignore if all the exception rules are matched rather than any one of them?

(I can open this as a separate issue if we'd like to discuss this as a general project principle.)

egibs: I've been thinking about that. We can workshop this 🙂

PR #1008, opened 2025-06-17
Address false positives with knative and new x86 third-party rule

egibs:
A couple of our new rules were rendering false positives for a handful of packages/dependencies.

The `CAPE_Formhookb` rule was wrong across every `.dll` and `.exe` file it detected (which I double-checked with VT).

The Knative findings were only a subset of the strings which is noisy (the problematic strings were not present). This is the problem with using `any of them` rather than doing something like `1 of $problem and any of $whatever_else`.


eslerm (approved):
Ack, but workarounds introduced should be addressed in future. cc @stevebeattie

PR #1018, opened 2025-06-27
Address false positives with dojo.js and YUI

egibs:
This PR contains false positive fixes for a handful of JS files. I scanned the files with these changes and they no longer show as critical.


eslerm (approved):
I researched these strings/wso2 earlier. Very eol protocol affecting legacy support in Win11. Not concerned :+1:

PR #1031, opened 2025-07-02
Add nil checks for slice pointers; filter invalid overrides from reports

egibs:
This PR adds guards for the new syscalls/pledges/caps slice pointers and finally refactors the override behavior to filter out matching rules that either reference invalid rule names or rules that were not part of the matching rules for a file.

This has the effect of removing seemingly-legitimate rule matches for a handful of samples, but these are the override rules that were matching (usually these would be prefixed with `false-positives/` but we had a handful of overrides scattered elsewhere).


pkg/report/report.go
```diff
@@ -656,14 +643,47 @@ func handleMetadata(m *yarax.Rule, b *malcontent.Behavior, fr *malcontent.FileRe
 			// YARAforge forgets to encode spaces
 			b.RuleURL = fixURL(v)
 		case "pledge":
-			*pledges = append(*pledges, v)
+			// pledges should not be nil when we get here, but guard against it
```
eslerm, line 646: 😎

eslerm (approved):

PR #1037, opened 2025-07-07
Drop user blocklist rule to high; add additional Bandit rules

egibs:
The Bandit username rule has always been much too noisy; this PR drops its risk to `high`, removes the negations, and supplements this with additional rules from the IOC list which are also categorized as `high`.


stevebeattie (approved):
LGTM, thanks!

PR #1045, opened 2025-07-14
reduce allocations in GetExt

Ritish134:
## Reduce allocations in `GetExt`

This change optimizes the `GetExt` function by replacing `strings.Split` and `fmt.Sprintf` with simpler index lookups and string slicing.  
This results in fewer heap allocations (`allocs/op`) when scanning large sets of files and slightly better performance.

---

###  Benchmark before:

Benchmark Before

---

###  Benchmark after:

Benchmark After

---


egibs (approved):
Nice! I wasn't a huge fan of my original implementation and this is much better.

PR #1055, opened 2025-07-24
Fix QuantityIncreasesRisk edge case with only high behaviors present

egibs:
@jamie-albert pointed out that aggregated high behaviors which register as critical would be dropped when using `--min-risk=critical` and `--min-file-risk=critical`.

This was mainly due to `handleOverrides` dropping any behaviors that weren't greater than or equal to the `minRisk` (4).

This PR fixes that edge case and adds additional guardrails when using `--quantity-increases-risk` so that we preserve any behaviors until we finally get to `if c.QuantityIncreasesRisk && upgradeRisk(ctx, overallRiskScore, riskCounts, size)...`.

I also de-duped the usage of `HighestMatchRisk` so we don't have to do that twice per file.

Verifying that this works:
```
$ out/mal --min-risk=critical --min-file-risk=critical --quantity-increases-risk=true scan ~/Downloads/fps/kibana-9-9.0.4-r0.apk
🔎 Scanning "~/Downloads/fps/kibana-9-9.0.4-r0.apk"
├─ 😈 ~/Downloads/fps/kibana-9-9.0.4-r0.apk ∴ /usr/share/kibana/node_modules/gpt-tokenizer/dist/o200k_base.js
│     • anti-behavior/blocklist/user — avoids execution if user has a particular name:
│     server, george, Louise, Lucas, Julia, Frank, fred, test, Lisa, John, Abby, mike
│     • exfil/stealer/creds — suspected data stealer: Bookmarks, Chromium, Telegram, Firefox, Binance, Discord, Bitcoin, History, Atomic, Chrome, Exodus
│     • impact/exploit — References 'explot'&{0xc000022180}
├─ 😈 ~/Downloads/fps/kibana-9-9.0.4-r0.apk ∴ /usr/share/kibana/node_modules/gpt-tokenizer/src/bpeRanks/o200k_base.js
│     • anti-behavior/blocklist/user — avoids execution if user has a particular name:
│     server, george, Louise, Lucas, Julia, Frank, fred, test, Lisa, John, Abby, mike
│     • exfil/stealer/creds — suspected data stealer: Bookmarks, Chromium, Telegram, Firefox, Binance, Discord, Bitcoin, History, Atomic, Chrome, Exodus
│     • impact/exploit — References 'explot'&{0xc000022180}
├─ 😈 ~/Downloads/fps/kibana-9-9.0.4-r0.apk ∴ /usr/share/kibana/node_modules/@kbn/security-solution-plugin/target/public/securitySolution.chunk.4588.js
│     • c2/addr/url — Contains HTTP hostname with unusual top-level domain: http://401trg.pw/, http://krbtgt.pw/
│     • c2/tool_transfer/download — References known file hosting site:
│     pastebin.com/m44e60e60, pastebin.com/NP64hTQr, pastebin.com/XAG1Hnfd, pastebin.com/qnLmpKuQ, pastebin.com/0dAciksC, pastebin.com/C0arvGxU, p…
│     • c2/tool_transfer/exe_url — accesses hardcoded powershell file endpoint: http://www.searchmiracle.com/silent.exe
│     • c2/tool_transfer/grayware — References websites that host code that can be used maliciously:
│     packetstormsecurity, pentestmonkey.net, exploit-db.com, milw0rm.com, 1337day.com
│     • impact/remote_access/reverse_shell — references a reverse shell: webshell
├─ 😈 ~/Downloads/fps/kibana-9-9.0.4-r0.apk ∴ /usr/share/kibana/node_modules/@kbn/fleet-plugin/target/public/fleet.chunk.720.js
│     • anti-static/obfuscation/bitwise — uses an excessive amount of unsigned bitwise math:
│     function(, charAt(r, charAt(i, charAt(a, charAt(n, charAt(t, a>>>13, a>>>26, l>>>26, o>>>24, t>>>26, d>>>13, h>>>13, m>>>13, b>>>13, k>>>13,…
│     • exec/remote_commands/code_eval — Likely evaluates encrypted content via fromCharCode: fromCharCode(55296+, fromCharCode(i, fromCharCode(e, eval(
│     • net/download/fetch — high-risk fetch command:
│     curl -L -O ${m}/beats/elastic-agent/elastic-agent-${l}-darwin-aarch64.tar.gz ${y}, curl -L -O ${a}/beats/elastic-agent/elastic-agent-${e}-da…

💡 For detailed analysis, try "mal analyze <path>"
```

### DIFF
--- a/pkg/report/report.go
+++ b/pkg/report/report.go
@@ -377,7 +377,7 @@ func fileMatchesRule(meta []yarax.Metadata, ext string) bool {
 	return true
 }
 
-func Generate(ctx context.Context, path string, mrs *yarax.ScanResults, c malcontent.Config, expath string, _ *clog.Logger, fc []byte, kind *programkind.FileType) (*malcontent.FileReport, error) {
+func Generate(ctx context.Context, path string, mrs *yarax.ScanResults, c malcontent.Config, expath string, _ *clog.Logger, fc []byte, kind *programkind.FileType, highestRisk int) (*malcontent.FileReport, error) {
 	if ctx.Err() != nil {
 		return &malcontent.FileReport{}, ctx.Err()
 	}
@@ -408,7 +408,6 @@ func Generate(ctx context.Context, path string, mrs *yarax.ScanResults, c malcon
 	risk := 0
 	riskCounts := make(map[int]int, 0)
 
-	highestRisk := HighestMatchRisk(mrs)
 	// Store match rules in a map for future override operations
 	mrsMap := createMatchRulesMap(mrs, matchCount)
 
@@ -435,9 +434,9 @@ func Generate(ctx context.Context, path string, mrs *yarax.ScanResults, c malcon
 		switch {
 		case risk == -1:
 			continue
-		case risk < minScore && !ignoreMalcontent && !override:
+		case !c.Scan && risk < minScore && !ignoreMalcontent && !override:
 			continue
-		case c.Scan && risk < highestRisk && !ignoreMalcontent && !override:
+		case c.Scan && risk < highestRisk && !c.QuantityIncreasesRisk && !ignoreMalcontent && !override:
 			continue
 		}
 
@@ -483,13 +482,15 @@ func Generate(ctx context.Context, path string, mrs *yarax.ScanResults, c malcon
 	}
 
 	// Update the behaviors to account for overrides
-	fr.Behaviors = handleOverrides(fr.Behaviors, fr.Overrides, minScore)
+	fr.Behaviors = handleOverrides(fr.Behaviors, fr.Overrides, minScore, c.Scan, c.QuantityIncreasesRisk)
 
 	// Adjust the overall risk if we deviated from overallRiskScore
 	// Scans will still need to drop <= medium results
-	newRisk := highestBehaviorRisk(fr)
-	if overallRiskScore != newRisk {
-		overallRiskScore = newRisk
+	overallRiskScore = highestBehaviorRisk(fr)
+
+	// If something has a lot of high, it's probably critical
+	if c.QuantityIncreasesRisk && upgradeRisk(ctx, overallRiskScore, riskCounts, size) {
+		overallRiskScore = 4
 	}
 
 	if c.Scan && overallRiskScore < HIGH {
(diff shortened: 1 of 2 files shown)


pkg/report/report.go
```diff
@@ -798,6 +792,11 @@ func handleOverrides(original, override []*malcontent.Behavior, minScore int) []
 
 	modified := make([]*malcontent.Behavior, 0, len(behaviorMap))
 	for _, b := range behaviorMap {
+		// if running a scan and using quantityIncreasesRisk,
+		// append every behavior so we can handle filtering correctly
+		if scan && quantityIncreasesRisk && b.RiskScore >= 3 {
```
stevebeattie, line 797:
```suggestion
		if scan && quantityIncreasesRisk && b.RiskScore >= HIGH {
```

stevebeattie: We have an enum defined in report.go, we should probably use it consistently, and ideally we'd define the levels globally somewhere, so that we don't have hardcoded integer values all over the place. Fixing that is not a blocker for this PR, though.

pkg/report/report.go
```diff
@@ -483,13 +482,15 @@ func Generate(ctx context.Context, path string, mrs *yarax.ScanResults, c malcon
...
 	// Adjust the overall risk if we deviated from overallRiskScore
 	// Scans will still need to drop <= medium results
-	newRisk := highestBehaviorRisk(fr)
-	if overallRiskScore != newRisk {
-		overallRiskScore = newRisk
+	overallRiskScore = highestBehaviorRisk(fr)
+
+	// If something has a lot of high, it's probably critical
+	if c.QuantityIncreasesRisk && upgradeRisk(ctx, overallRiskScore, riskCounts, size) {
+		overallRiskScore = 4
```
stevebeattie, line 493:
```suggestion
		overallRiskScore = CRITICAL
```

stevebeattie (changes requested):
There's a lot of twisty logic in this, based on what command line arguments are passed. It'd be nice if we had some tests around this so that we make sure we're handling this logic correctly.

stevebeattie (approved):
Okay, I think I understand the logic and the changes, and am approving the PR. Would still like to capture the desired behaviors in some test cases.

Thanks!

PR #1078, opened 2025-08-09
Tune concurrency and scanFD

egibs:
This PR makes some minor tweaks to how per-scan concurrency is handled along with the `scanFD` method which was duplicating size calculations.

While it shouldn't matter too much, we can cap the concurrency for processing standard file paths along with extracted file paths so that we don't just use the configured concurrency value everywhere.

We still need to create a 1:1 pool of scanners to available CPUs since we use the pool for the duration of a `mal` run which may execute concurrent scans of single files (like `make refresh-sample-testdata`).


pkg/action/scan.go
```diff
@@ -118,22 +103,27 @@ func scanSinglePath(ctx context.Context, c malcontent.Config, path string, ruleF
...
-	defer f.Close()
 
 	fi, err := f.Stat()
 	if err != nil {
 		return nil, err
 	}
 
 	size := fi.Size()
-	if size == 0 {
+	if size <= 0 {
```
eslerm, line 113: why can size be less than zero?

egibs: Updated in [`95e359c` (#1078)](https://github.com/chainguard-dev/malcontent/pull/1078/commits/95e359c9b9aa9272519d09651b36db8b2b586e26).

eslerm (approved):

eslerm (approved):

PR #1139, opened 2025-09-26
Clean up some code and consolidate variables

egibs:
This PR consolidates some of the `diff.go` variables since we do the same declarations for both the source and destination variables. We also never used `action.go` and there were a couple of cases where we increased complexity of `if` statements when we could just have an earlier, separate check to return early.

`refresh.go` was still using `filepath.Walk` instead of `WalkDir` with the latter being much more efficient even though we aren't enumerating thousands of paths when discovering test data.

These changes shouldn't have any functional impact (i.e., `make refresh-sample-testdata` is a NOP).


pkg/action/diff.go
```diff
@@ -171,29 +178,21 @@ func relFileReport(ctx context.Context, c malcontent.Config, fromPath string, is
...
 
 	for _, pattern := range patterns {
 		re := regexp.MustCompile(pattern)
-		if re.MatchString(fr.Path) {
-			scoreSrc = true
-		}
-		if re.MatchString(tr.Path) {
-			scoreDest = true
+
+		// If both files match patterns, reeturn true to indicate that `inferMoves` should be used
```
egibs, line 189:
```suggestion
		// If both files match patterns, return true to indicate that `inferMoves` should be used
```

PR #1204, opened 2025-11-07
Add fuzzing for extraction, file type determination, and report generation

egibs:
This PR adds fuzzing for the core functionality around extracting archives, determining file types, and generating reports and also contains a fix/update for the `IsValidPath` function which the fuzzing exposed.


eslerm (approved):
approved with non-blocking comments

PR #1209, opened 2025-11-10
Handle file contents/checksum only when generating a report

egibs:
This PR updates the retrieval of a file's contents and the calculation of its checksum to only occur if we're going to generate a report. Otherwise, we're wasting resources on files that won't be displayed as part of the final results.


stevebeattie (approved):
LGTM, thanks!

Issue #1230, opened 2025-11-23
Diff doesn't work as expected


Ais8Ooz8:
As far as I understand, diff should be zero for identical files or for files without significant changes. But my tests with trivy binaries show rather strange results. I'm testing on MacOS, installed it with `brew install malcontent`, and `mal --version` shows `malcontent version v1.17.4`.
```
wget -q https://github.com/aquasecurity/trivy/releases/download/v0.67.1/trivy_0.67.1_Linux-64bit.tar.gz
```
```
wget -q https://github.com/aquasecurity/trivy/releases/download/v0.67.2/trivy_0.67.2_Linux-64bit.tar.gz
```

```
mal diff 0.67.1/trivy 0.67.1/trivy | tee >(sha256sum) >(wc -l) > /dev/null
     249
47b537d7255958967449a6419e09500e4f1a8f7015ea523b5895bc3979cc8b10  -

mal diff 0.67.1/trivy 0.67.2/trivy | tee >(sha256sum) >(wc -l) > /dev/null
     249
47b537d7255958967449a6419e09500e4f1a8f7015ea523b5895bc3979cc8b10  -

mal diff 0.67.2/trivy 0.67.1/trivy | tee >(sha256sum) >(wc -l) > /dev/null
     249
47b537d7255958967449a6419e09500e4f1a8f7015ea523b5895bc3979cc8b10  -

mal diff 0.67.2/trivy 0.67.2/trivy | tee >(sha256sum) >(wc -l) > /dev/null
     249
47b537d7255958967449a6419e09500e4f1a8f7015ea523b5895bc3979cc8b10  -
```

```
mal diff 0.67.2/trivy 0.67.2/trivy_upx | tee >(sha256sum) >(grep +) > /dev/null
47b537d7255958967449a6419e09500e4f1a8f7015ea523b5895bc3979cc8b10  -

mal diff 0.67.2/trivy 0.67.2/trivy_upx | tee >(sha256sum) >(grep +) > /dev/null
│+      🛑 packer/upx — Binary is packed with UPX: This file is packed, executable packer, UPX!
│+      🟡 embedded/zstd — Contains compressed content in ZStandard format: $ref
│+      🟡 bypass_security/linux/ufw — interacts with the ufw firewall: disable, allow, deny
│+      🟡 path/dev — path reference within /dev: /dev/shm
641996f987c08f62ecf690115a60a5b8027c58380f76ed70cc1b8f3e4a0b85d3  -

[...]

PR #1234, opened 2025-11-24
feat: simplify file content retrieval and address diff inconsistencies

egibs:
Relates to: #1230

`programkind.File` originally considered only the first 512 bytes of a file's header. When scanning UPX files, we were ignoring legitimate UPX files (as well as other file types) and not decompressing them which would appear oddly in diffs (see the aforementioned Issue).

Since we were already reading file contents in at least two places, this PR consolidates file content retrieval into just `programkind.File` and returns the results if we successfully determine a file's type.

The only overhead with this approach is that we now persist a file's contents even if we don't end up generating a report for it; however, this allows us to more correctly identify all files.

We also preserved the original UPX file which duplicated scan results in a confusing way. This PR also does away with the `-k` flag so that we only scan the decompressed file. We also displayed the entire diff report when there were no changes which was also confusing.

With these changes, things look a little better. For example, diffing a non-UPX file and a UPX file with the latter being successfully decompressed produces a diff with no changes:
```
$ out/mal diff out/lib/libyara_x_capi.so.1.9.0 out/lib/libyara_x_capi.so.1.9.0_upx

```

The non-UPX improvements can also be seen in the newly-generated results.


pkg/action/scan.go
```diff
@@ -139,31 +128,24 @@ func scanSinglePath(ctx context.Context, c malcontent.Config, path string, ruleF
...
 		return nil, NewFileReportError(err, path, TypeGenerateError)
 	}
 
 	defer func() {
 		f.Close()
-		readPool.Put(buf)
 		scannerPool.Put(scanner)
 		mrs = nil
-		fc.Reset()
+		fc = fc[:0]
```
eslerm: The doesn't clear the data (?)

PR #1501, opened 2026-05-04
Detect npm credential exfil in install scripts

alephnull-sh:
Adds a high-risk npm rule for install-time scripts that touch package-manager credentials and also send data over HTTP.

I added a small fixture/test with a `postinstall` script that reads `.npmrc` / token env vars and posts them, plus a release-script fixture that mentions `NPM_TOKEN` but should stay quiet.

The rule metadata also links Unit 42's [Shai-Hulud writeup](https://unit42.paloaltonetworks.com/npm-supply-chain-attack/), since that attack used npm install hooks to harvest and exfiltrate developer credentials.

I also opened chainguard-sandbox/malcontent-samples#52 so this can be covered by the sample-output tests once that sample lands.

Checked:
- `make yara-x-compile`
- local YARA-X scan against the positive and negative fixtures


egibs (approved):
Thanks for the contribution!

